PR #240, opened 2016-04-14
handle sparqlUpdate request as single string arg

### LINKED ISSUE #236: unable to invoke sparqlUpdate with just update string param
This might be a jsdoc bug or user head gap, but the JSDoc for graphs.sparqlUpdate implies (well, more than implies) you can just pass a string as input: "Executes a SPARQL update against the graphs; takes a configuration object with the following named parameters or, **as a shortcut, the SPARQL update data**."

However, I find this is not so. For example, this works:
```javascript
db.graphs.sparqlUpdate({
   data: 'CREATE GRAPH <http://marklogic.com/semantics/tutorial/update> ;'
 })
```
But this does not:
```javascript
db.graphs.sparqlUpdate('CREATE GRAPH <http://marklogic.com/semantics/tutorial/update> ;')
```
The latter coughs up the following error:
```
"RESTAPI-INVALIDCONTENT: (err:FOER0000) Invalid content: No query to execute"
```

I'll tag this for 8.0-5, but it's probably too trivial to be worth worrying your pretty little head over at this point.

wooldridge:
This bug fix supports the shortcut way of calling graph.sparqlUpdate(). The method can also be called with a configuration object.

closes #236

### DIFF
--- a/lib/graphs.js
+++ b/lib/graphs.js
@@ -787,8 +787,10 @@ Graphs.prototype.sparqlUpdate = function updateGraphSPARQL() {
     defaultRulesets    = arg.defaultRulesets;
     optimizeLevel      = arg.optimizeLevel;
     bindings           = arg.bindings;
-  } else {
-    data = args[1];
+  }
+  // SPARQL request is single string argument
+  else {
+    data = args[0];
   }
 
   var endpoint = '/v1/graphs/sparql';


ehennum:
A minor nit:
    data = arg; 
would also work and is consistent with the pattern elsewhere, but the difference is fussy and of no account.

Thanks for squaring this issue away.

PR #241, opened 2016-04-15
correctly handle default graphs in multi-param sparql()

### LINKED ISSUE #237: potential for graphs.sparql to drop default graph uri
Rummaging around in the impl for graphs.sparql, I noticed what may be a problem with the parameter handling if you pass in only a single default graph URI. (It could also be my meager JS knowledge leading me astray. You have been warned!)

I think there's an off by 1 error in this block of code at  https://github.com/marklogic/node-client-api/blob/master/lib/graphs.js#L629 that handles the no-call-object pattern.
```javascript
    acceptType = args[0];
    if (argLen > 3) {                // <-- DANGER, WILL ROBINSON
      defaultGraphs = Array.prototype.slice.call(arguments, 1, argLen - 1);
    }
    query = args[argLen - 1];
```
Isn't that going to do the wrong thing if there is exactly one default graph uri? That is, it seems like if I invoke graphs.sparql like this, it won't pick up the graph uri because argLen will be exactly 3.

    db.graphs.sparql(contentType, oneLonelyUri, query)
AFAICT, the unit tests never try this form of call. I confess, I haven't tried this myself because SPARQL terrifies me. Default what? I wouldn't know if it worked or not.

wooldridge:
closes #237

### DIFF
--- a/lib/graphs.js
+++ b/lib/graphs.js
@@ -607,6 +607,7 @@ Graphs.prototype.sparql = function queryGraphSPARQL() {
 
   // TODO: collection, directory?
 
+  // Single configuration object
   if (argLen === 1) {
     var params = args[0];
     acceptType    = params.contentType;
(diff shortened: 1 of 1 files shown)


ehennum:
Thanks, Kim and Mike, for catching and correcting that howler.

PR #242, opened 2016-04-15
remove unnecessary code in requester.js

wooldridge:
authenticatedRequest() is called twice, and neither call uses a return value, so safe to remove "return request".

operation.requestBodyProvider property is never set, so safe to remove the logic that acts on it.


ehennum:
Mike, thanks for cleaning up the code base.  

I filled in the labels, milestone, and assignee.

Aries, this refactoring requires no new tests -- it simply just shouldn't cause any regressions.

PR #243, opened 2016-05-09
Replace mlutil.callbackOn() with func.bind()

wooldridge:
The standard built-in function provides the same functionality.

### DIFF
--- a/etc/test-setup-prompt.js
+++ b/etc/test-setup-prompt.js
@@ -22,7 +22,7 @@ function adminUserPrompt() {
     read({
       prompt: 'admin user (default=admin): '
       },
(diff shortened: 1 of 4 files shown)


ehennum:
Hi, Mike:

Ordinarily, I would also expect built-in operations to be more performant as well as cleaner, but it appears that bind() has a fundamental design issue:
 
http://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure

Unless we have some confirmation that v8 has come up with a way to make bind() as fast as a closure, maybe we should continue to use closures.

If that sounds right, please close the issue.

wooldridge:
Thanks for the explanation, Erik.

PR #265, opened 2016-07-26
Added support for metadata values

### LINKED ISSUE #259: Add support for metadata values
Support the new name-value string pairs called metadata just like we already support document properties, quality, permissions, and collections.

wooldridge:
Fixes #259

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -131,6 +131,7 @@ function probeOutputTransform(/*headers, data*/) {
  * write the document
  * @property {object[]} [properties] - additional properties of the document
  * @property {number} [quality] - a weight to increase or decrease the rank of the document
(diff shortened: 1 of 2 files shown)


lib/documents.js
```diff
@@ -340,6 +341,13 @@ function readDocumentsImpl(contentOnly, args) {
     categories = [categories];
   }
 
+  if (categories != null) {
+    var i = 0;
+    for (i = 0; i < categories.length; i++) {
+      categories[i] = categories[i] === 'metadataValues' ? 'metadata-values' : categories[i];
+    }
+  }
+
```
ehennum: Should the function copy instead of modifying the caller's array?  Possibly a mapper function would be a straightforward way to make the change at the same time as copying?

PR #267, opened 2016-07-26
Add support for Kerberos authentication

### LINKED ISSUE #260: Add support for Kerberos authentication
Support Kerberos single sign-on security when it is applied to MarkLogic application servers.

wooldridge:
Fixes #260

### DIFF
--- a/lib/requester.js
+++ b/lib/requester.js
@@ -15,10 +15,13 @@
  */
 'use strict';
 var createAuthInitializer = require('www-authenticate');
+var Kerberos              = require('kerberos').Kerberos;
 var Multipart             = require('multipart-stream');
 var through2              = require('through2');
 var mlutil                = require('./mlutil.js');
 var responder             = require('./responder.js');
+
+var kerberos = null;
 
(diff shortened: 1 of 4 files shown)


lib/requester.js
```diff
@@ -140,6 +171,32 @@ function challengeRequest(operation, requester) {
   request1.on('error', operation.errorListener);
   request1.end();
 }
+function credentialsRequest(operation, requester) {
+  kerberos = new Kerberos();
```
ehennum: Is it necessary to initialize the kerberos variable any time that credentialsRequest() is executed? Probably so, but if not, maybe it would be better to initialize only if the variable is null.

lib/requester.js
```diff
@@ -140,6 +171,32 @@ function challengeRequest(operation, requester) {
...
+  var uri = "HTTP@"+operation.options.host;
+  kerberos.authGSSClientInit(uri, 0, function(err, ctx) {
+    if (err) {
+      operation.errorListener('kerberos initialization failed at '+uri);
```
ehennum: Should the function return at this point instead of falling through?

Issue #275, opened 2016-08-11
Move kerberos dependencies to optionalDependencies in package.json


wooldridge:
Making kerberos and related modules optional is useful for developers who don't need Kerberos authentication (the majority don't) and don't have the environment to support it. More info:

https://docs.npmjs.com/files/package.json#optionaldependencies

The kerberos module includes native C++ code and is a more complicated dependency to get working, more info here:

https://www.npmjs.com/package/kerberos

@kcoleman-marklogic brought this issue to our attention.

@ayuwono adds:

There are several dependencies need to be installed on the machine, I can name:

Node-gyp
Python
Glibc-devel 64-bit (sudo yum install glibc-devel.x86_64)

And I think it has different dependencies on windows machine too

kcoleman-marklogic:
This might end up being a useful reference to include in the README, as it does talk about the deps. Silly me, not to have looked there previously.

https://www.npmjs.com/package/kerberos

kcoleman-marklogic:
I have officially given up on getting Node.js to npm install on Windows until Kerberos slinks off to the corner where it belongs. 

Heaven help the poor user who actually needs our API + Kerberos on Windows. I've wasted most of this week trying to get our install working on Windows, without success. I wound up in Windows SDK hell, and am now worse off than when I started. 

We won't have updated JSDoc for the API on pubs until this issue is addressed, which I realize is a minor matter compared to the potential user pain. I mention it only in case you go looking for the docs and wonder why they are not up to date.

wooldridge:
@kcoleman-marklogic hoping the #277 merge addresses your install issues. @ehennum thanks for the help!

kcoleman-marklogic:
Here is what I found:

`npm install` coughs up a bunch of errors because it still tries to install the optional deps by default. However, the install succeeds now since the failed dep is optional.

`npm install --no-optional` completes without error since it skips over the Kerberos stuff altogether.

If that's the desired usage, then we're good to go. I do think it is potentially alarming/confusing to see those errors, even though the install succeeds. It looks just like when it fails. :-/ It'd be nice if Kerberos was somehow opt-in instead of just skippable, but I'm not sure npm puts those tools in our hands.

kcoleman-marklogic:
P.S. I also made another attempt to get the kerberos install working today, just because it bugs me to be defeated by an installer, but I was beaten again. I got node-gyp to install finally, but Kerberos still won't build.

[...]

PR #277, opened 2016-08-12
Make kerberos module an optional dependency

wooldridge:


wooldridge:
Partly addresses issue #275, still need documentation to describe all dependencies for running Kerberos.

PR #286, opened 2016-08-30
Allow custom http(s) agent to be used

elliotttf:
Currently a default http(s) agent is provided via Yakaa if the `agent` configuration option is not set. If it is set though the agent is completely omitted from the `connectionParams` object. This change will maintain existing behavior of providing a default agent but also allows users to specify one of their own.

### DIFF
--- a/lib/marklogic.js
+++ b/lib/marklogic.js
@@ -621,10 +621,16 @@ function initClient(client, inputParams) {
       });
       connectionParams.agent = new YAgent.SSL(agentOptions);
     }
+    else {
+      connectionParams.agent = inputParams.agent;
+    }
   } else {
     client.request = http.request;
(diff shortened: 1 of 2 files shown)


jmakeig:
@elliotttf, thanks for the PR. Before we can consider your contribution, though, our lawyers need a signed [CLA](http://developer.marklogic.com/products/cla) on file. (You’ll only have to do that once.)

Again, thanks and let me or @wooldridge know if you have questions or concerns.

elliotttf:
@jmakeig @wooldridge sent, let me know if you need anything else in that department, otherwise I'm happy to respond to any questions or feedback for this review.

wooldridge:
These updates have been cherrypicked onto the develop and 1.0-develop branches.

PR #287, opened 2016-08-31
Use var instead of const in variable declaration

wooldridge:
ES6 features cause issues with earlier versions of Node.js.


wooldridge:
@ayuwono as we discussed regarding test setup failure

PR #288, opened 2016-09-01
Remove gulp-jsdoc dep from package.json until taffydb issue is fixed

wooldridge:
A renaming has broken apps that depend on gulp-jsdoc:
https://github.com/hegemonic/taffydb/pull/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "bunyan": "^1.3.3",
     "core-util-is": "^1.0.1",
     "gulp": "^3.8.10",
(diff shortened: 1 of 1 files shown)


kcoleman-marklogic:
Do you have to remove it, or can we change the dep to some older version? I'm concerned that if you just yank it out, we will no longer be able to build the JSDoc for the project.

kcoleman-marklogic:
This is also likely to break the nightly doc builds, unless I misunderstand what's going on.

PR #308, opened 2016-10-27
Prevent premature socket destruction

elliotttf:
Calling `response.destroy()` on an empty response leads to the underlaying socket to be closed. This is probably desirable in an error case, but if the response is expected to be empty it results in socket closures even when socket pooling is enabled.
## Steps to reproduce
1. Configure MarkLogic to use a connection pool, e.g.
   
   ``` javascript
   const marklogic = require('marklogic');
   const Agent = require('yakaa');
   
   const agent = new Agent({
   kepAlive: true,
   keepAliveMsecs: 1000,
   keepAliveTimeoutMsecs: 15000,
   maxSockets: 42,
   maxFreeSockets: 10,
   });
   
   const db = marklogic.createDatabase({
   agent,
   host: 'ml.example.com',
   port: 9999,
   user: 'sample_user',
   password: 'sample_password'
   });
   ```
2. Issue a call to MarkLogic that results in an expected empty response – specifically I observed this when saving documents with a given URI that did not already exist in the database
   
   ``` javascript
   db.documents.write([
   {
     uri: 'http://data/ns.org.com/entityType/new-entity-id.json',
     content: {
       // ...
     }
   }
   ])
   .result()
   ```
3. Observe that a `PUT` is issued to the server with an empty response

The last observation _may_ be a result of the way the MarkLogic instance I'm querying is configured, but regardless, the empty response causing the socket to be destroyed nullifies all benefits of connection pooling.

### DIFF
--- a/lib/requester.js
+++ b/lib/requester.js
@@ -126,7 +126,6 @@ function challengeRequest(operation, requester) {
       var authenticator = (hasChallenge) ? createAuthenticator(
           operation.client, options.user, options.password, challenge
           ) : null;
-      response1.destroy();
       authenticatedRequest(operation, authenticator, requester);
     // should never happen
     } else if (successStatus && isRead) {
--- a/lib/responder.js
+++ b/lib/responder.js
@@ -32,7 +32,6 @@ function responseDispatcher(response) {
 
   var outputMode = operation.outputMode;
   if (outputMode === 'none') {
-    response.destroy();
     return;
   }
 
@@ -83,7 +82,6 @@ function responseDispatcher(response) {
   if (isEmpty) {
     if (expectedType !== 'empty') {
       operation.logger.debug('expected body or multipart but received empty response');
-      response.destroy();
     }
 
     if (outputMode === 'promise') {
@@ -778,7 +776,6 @@ function isResponseStatusOkay(response) {
           }));
     } else {
       operation.errorListener(clientError);
-      response.destroy();
     }
 
     return false;
--- a/lib/responder.js
+++ b/lib/responder.js
@@ -83,9 +83,8 @@ function responseDispatcher(response) {
   if (isEmpty) {
     if (expectedType !== 'empty') {
       operation.logger.debug('expected body or multipart but received empty response');
+      response.destroy();
     }
-
-    response.destroy();
 
     if (outputMode === 'promise') {
       dispatcher.emptyPromise(response);


ehennum:
Hi, Eliott:

Thanks heaps for pointing out this problem.

First off, we have a few preliminaries for contributions:

https://github.com/marklogic/node-client-api/blob/master/CONTRIBUTING.md
 
Second, I have a question about the patch. 

While it's a good point that the client should not destroy the empty response, I suspect that the client needs to read the empty response to close it (which would be better than destroy() even in the case where the warning is logged).  

I haven't tried, but it might work to use concat-stream and throw away the result.

The response module has two other calls to destroy() that probably need similar treatment.

@wooldridge, what do you think?

elliotttf:
@ehennum thanks, I've already done the CLA (see #286) and I've updated the description with steps to reproduce the problem. I noticed that the code I was changing was not directly covered in existing test cases (or maybe I missed it?) so I didn't add a test at this moment in favor of getting the discussion started.

I'll do some more debugging but I think the response was still being closed – it's possible it was just timing out eventually though. I'll check on the context of the other `destroy` cases while I'm in there, they may be valid.

ehennum:
Ach, of course you have done the CLA.  

Thanks for digging deeper into the issue -- greatly appreciated.

elliotttf:
@ehennum ok, after some further investigation, the sockets are correctly closing even without destroy. Per the [documentation](https://nodejs.org/docs/latest/api/net.html#net_socket_destroy_exception) destroy really only needs to be used in error cases.

Should the [destroy I changed](https://github.com/marklogic/node-client-api/pull/308/files#diff-6a57e93aad1925f27c57f4980375cae5R86) be treated like an error case or would it be better to remove it entirely?

The other destroys in the codebase are similar, only [this one](https://github.com/marklogic/node-client-api/blob/66cfbed180d6a85daf720c7a3040af66460f17da/lib/responder.js#L782) is an error case but even that is an HTTP error, not some error that would prevent future use of the socket so I think it can also be removed.

elliotttf:
4cb274f removes all `destroy()` calls but if you guys think they should still be used in the responder error cases I'll happily push another change for this.

ehennum:
@elliotttf , thanks for the diligence on this issue.  I'm persuaded (including that the socket doesn't have to be closed down in the error cases either).

@wooldridge , any concerns?

wooldridge:
@elliotttf thanks so much for the PR and the investigation. This fix makes sense given everything said.

@ayuwono let's discuss test coverage in this area.

wooldridge:
These updates have been cherrypicked onto the develop and 1.0-develop branches.

Issue #295, opened 2016-09-13
Add support for node-update


wooldridge:
node-update is a new document-level capability to support the element-level security feature. Implementing in the Node API is dependent on work at the REST layer:

https://bugtrack.marklogic.com/42097

wooldridge:
The API should support the new "node-update" capability as-is.

ayuwono:
Added test on node-update

PR #309, opened 2016-10-27
Test for "node-update" document-level capability

wooldridge:
Ensure that the new capability is covered in test-basic.


wooldridge:
Fixes #295

Issue #332, opened 2016-12-08
remove crosses, equals, and touches geo region operators from JSDoc


kcoleman-marklogic:
The geo region query operators crosses, equals, and touches have been dropped from ML9. They need to be removed from the list of available operator values in the JSDoc for the `operator` param of  `queryBuilder.geospatialRegion`.

http://pubs.marklogic.com:8011/9.0/jsdoc/queryBuilder.html#geospatialRegion

I'm going to assign this to you for now, Mike. If I get around to it first, I'll steal it from you.

wooldridge:
Thanks @kcoleman-marklogic, I will take care of it:

https://github.com/wooldridge/node-client-api/tree/ea4/issue-322-regiondoc

PR #338, opened 2016-12-20
edit docs to remove unsupported spatial operators

wooldridge:
the operators crosses, equals, and touches are not supported


wooldridge:
Fixes #332

Issue #335, opened 2016-12-08
Update docs for near()


wooldridge:
Need to say that indexes or filtering must be enabled for near() queries to work. Proposed:

"A near query requires the word positions index for the database to be turned on or the query to be run filtered by setting the search parameter in {@link queryBuilder#withOptions}. Otherwise, incorrect results may be returned. While turning on filtering can be convenient during development, setting up indexes is recommended for production."

Goes here:

https://docs.marklogic.com/jsdoc/queryBuilder.html#near

kcoleman-marklogic:
So...I asked Mary about the whole "requires word positions" thing because we don't say that in other places  (like the apidoc for `cts:near-query`, for example), and I wondered if we should. I got back a torrent of information, which I am still trying to absorb the gist of. I'll share it with you via email so you can contemplate what, if any, implications it has for this issue.

wooldridge:
Revised text based on chats with Kim (and input from Mary):

"Note that you may get false positives from a near query if you do not have the appropriate position index enabled or run the query unfiltered (which is the default). While turning on filtering can be convenient during development, setting up indexes is recommended for production."

PR #339, opened 2016-12-20
update docs for near() requirements

wooldridge:
To avoid false positives, you can turn on the word positions index or enable filtering


wooldridge:
Fixes #335

PR #352, opened 2017-01-30
fix(#250): add borrows JSDoc references to values-builder.js

### LINKED ISSUE #250: Documentation for valuesBuilder.geospatial() missing
Need to add a JSDoc "borrows" entry in values-builder.js for queryBuilder#geospatial:

http://usejsdoc.org/tags-borrows.html

https://github.com/marklogic/node-client-api/blob/master/lib/values-builder.js#L71

wooldridge:
For geospatial() and geospatialRegion().

Fixes #250

### DIFF
--- a/lib/values-builder.js
+++ b/lib/values-builder.js
@@ -441,6 +441,7 @@ module.exports = {
     geoPropertyPair:    qb.geoPropertyPair,
     geoOption:          qb.geoOption,
     geospatial:         qb.geospatial,
+    geospatialRegion:   qb.geospatialRegion,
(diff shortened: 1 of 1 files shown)


wooldridge:
@kcoleman-marklogic can you review this fix?

kcoleman-marklogic:
I would be glad to, if I had any idea what the result is supposed to be. The issue (#250) describes the solution, not the problem.

kcoleman-marklogic:
This looks OK to me, modulo what I said about region lexicon queries in #250. I think you need to take out the `geospatialRegion` ref.

Issue #255, opened 2016-07-09
db.documents.probe() throws error when transaction Id is passed as second param


csreddy:
the documentation shows that `probe()` takes 2 params `uri` and `txid` but its throws below error
ex:

``` javascript
 db.documents.probe('/test/transaction/doc1.json', tid).result();
```

works only if passed as an object

``` javascript
 db.documents.probe({uri:'/test/transaction/doc1.json', txid: tid}).result();
```

```
 Uncaught Error: must supply uri for document check()
      at Documents.probeDocumentsImpl (/Users/sreddy/space/angularjs/node-client-api/lib/documents.js:171:11)
      at Documents.probeDocument [as probe] (/Users/sreddy/space/angularjs/node-client-api/lib/documents.js:166:29)
      at /Users/sreddy/space/angularjs/node-client-api/test-complete/nodejs-probe-txn.js:34:27
      at tryCatcher (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/util.js:24:31)
      at Promise._settlePromiseFromHandler (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/promise.js:454:31)
      at Promise._settlePromiseAt (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/promise.js:530:18)
      at Promise._settlePromises (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/promise.js:646:14)
      at Async._drainQueue (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/async.js:177:16)
      at Async._drainQueues (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/async.js:187:10)
      at Immediate.Async.drainQueues [as _onImmediate] (/Users/sreddy/space/angularjs/node-client-api/node_modules/bluebird/js/main/async.js:15:14)
      at tryOnImmediate (timers.js:534:15)
      at processImmediate [as _immediateCallback] (timers.js:514:5)
```

PR #356, opened 2017-02-02
fix(#255): enable probe() to accept params as list or object

wooldridge:
Includes tests for probe().


wooldridge:
@ayuwono, this fixes documents.probe() so that it can accept params as a list or an object. It also adds tests  for probe() (including tests using transactions) to test-basic. Can you review?

ayuwono:
Looks good to me Mike

Issue #231, opened 2016-02-02
opt into or remove deprecated interfaces


ehennum:
The deprecated interfaces include the following.

document.remove() returns a response with both a uris property and a uri property.  The singular property is a legacy from when document.remove() could only remove a single document.  We should remove that property in MarkLogic 9.

In MarkLogic 8, the developer has to opt in to get a transaction object and receives a transaction id string by default.  MarkLogic 9 should reverse the default and opt in.

In MarkLogic 8, the developer has to opt in for a zero-based slice on the search response and supplies a 1-based slice by default.  MarkLogic 9 should reverse the default and opt in.

wooldridge:
MarkLogicClient.remove() should also be addressed per the documentation:

> For backward compatibility, if a string is passed, the uri is returned as a string, but in the next major release, the uri will be return as an array with one string.

kcoleman-marklogic:
Mike, if you find other things to remove/change besides the items called out here, please add them so that I can ensure they all make it into the release notes.

wooldridge:
Per email discussion with @ehennum and @jmakeig, we will deprecate the transactionId string and legacy slice option in ML9 and remove those features entirely in ML10. I will add to the JSDoc regarding this in my check-in and @kcoleman-marklogic can review.

wooldridge:
@ayuwono, these updates to address deprecated features will break tests:

- Remove() methods for documents now only return arrays of one or more URIs. So any tests that expect a string return value need to be fixed.
- The transactions.open() method now returns a full-fledged Transaction object by default instead of an object with a transactionId string. So any tests not expecting the Transaction object will need to be fixed.
- The slice() method now works in "array" mode by default. This differs from "legacy" mode in how the params work. Tests that are working with sliced results will need to have their params updated.

ayuwono:
Changed all of the affected tests

wooldridge:
@kcoleman-marklogic, can you review the changes to the jsdoc comments for this fix? You can see what changed here:

https://github.com/marklogic/node-client-api/pull/360/files

Namely, there are comment changes for:

DatabaseClient.remove()
queryBuilder.slice()
transactions.open()

kcoleman-marklogic:
Bit late for me to mention this, I know, but shouldn't the slice change apply to values queries, too? It seems rather odd to only make it zero-based for search results.

kcoleman-marklogic:
[...]

PR #360, opened 2017-02-04
fix(#231): address deprecated features

wooldridge:
- legacy non-array return values for remove() methods
- default return values for transactions.open() (change from transaction ID to object)
- default slice mode (change from "legacy" to "array")


wooldridge:
Fixes #231

Issue #343, opened 2017-01-14
Outdated dependencies


jmakeig:
As of b8c7828bcac093aca50a78b506f0189c19551fb7, `npm outdated` yields:

```
Package           Current  Wanted  Latest  Location
bluebird           2.11.0  2.11.0   3.4.7  marklogic
concat-stream       1.5.2   1.6.0   1.6.0  marklogic
deepcopy            0.4.0   0.4.0   0.6.3  marklogic
gulp-jshint        1.12.0  1.12.0   2.0.4  marklogic
gulp-mocha          2.2.0   2.2.0   3.0.1  marklogic
kerberos           0.0.21  0.0.21  0.0.22  marklogic
mocha               2.5.3   2.5.3   3.2.0  marklogic
multipart-stream    1.0.0   1.0.0   2.0.1  marklogic
qs                  2.4.2   2.4.2   6.3.0  marklogic
should              4.6.5   4.6.5  11.1.2  marklogic
through2            0.6.5   0.6.5   2.0.3  marklogic
winston             0.9.0   0.9.0   2.3.0  marklogic
```

A major release is probably the best time to upgrade any/all of these. However, we should only upgrade if there's a good reason. Some of these may have incompatibilities.

jmakeig:
There are also a bunch of deprecation warnings on `npm install`:

```
npm WARN deprecated gulp-jsdoc@0.1.5: This project is deprecated. Use gulp-jsdoc3 instead.
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
npm WARN deprecated wrench@1.5.9: wrench.js is deprecated! You should check out fs-extra (https://github.com/jprichardson/node-fs-extra) for any operations you were using wrench for. Thanks for all the usage over the years.
npm WARN deprecated wrench@1.3.9: wrench.js is deprecated! You should check out fs-extra (https://github.com/jprichardson/node-fs-extra) for any operations you were using wrench for. Thanks for all the usage over the years.
[...]

PR #365, opened 2017-02-07
task(#343): update dependencies in package.json

wooldridge:
Updated dependencies to current versions based on running "npm outdated".
Updated file path for new version of bluebird.
Fixed minor linting errors.


wooldridge:
Fixes https://github.com/marklogic/node-client-api/issues/343

PR #369, opened 2017-02-14
fix(#227): clarify when a summary object appears in query results

### LINKED ISSUE #227: debug info is returned even when debug flag is set to false
Pardon me if i'm doing something wrong here. I expect debug data not to be returned when `debug` flag is set to false. When I remove `withOptions`, only then it would not return debug data, the boolean flag does not seem to take effect.

``` javascript
var marklogic = require('marklogic');

var db = marklogic.createDatabaseClient({
    host: 'staging-bugtrack.marklogic.com',
    port: 8006,
    user: 'admin',
    password: 'admin',
    authType: 'DIGEST'
});
var q = marklogic.queryBuilder;

var processedUris = [];

db.documents.query(q.where(
        q.directory('/b8_0/2015-11-01/', true)
    ).slice(1, 1)
    .withOptions({
        debug: false,
        queryPlan: false,
        metrics: false
   })
)
    .stream()
    .on('data', function(document) {
        console.log(document);
        processedUris.push(document.uri);
    })
    .on('error', function(error) {
        console.error(error);
        // done(error);
    })
    .on('end', function() {
        console.log('done ' + processedUris.length);
        console.log(processedUris);
    });
```

wooldridge:
Fixes #227

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -169,7 +169,7 @@ function wipeOutputTransform(/*headers, data*/) {
 /**
  * Categories of information to read or write for documents.
  * The possible values of the enumeration are
- * content|collections|permissions|properties|quality|metadataValues|metadata where
+ * content|collections|metadataValues|permissions|properties|quality|metadata|none where
  * metadata is an alias for all of the categories other than content.
  * @typedef {enum} documents.categories
  */
--- a/lib/query-builder.js
+++ b/lib/query-builder.js
@@ -4986,6 +4986,7 @@ function snippet() {
  * This function may be called on the result of building a query. When the
  * 'debug', 'metrics', 'queryPlan', or 'similarDocs' parameter is set, a
  * search summary object will be returned along with the result documents.
+ * When 'categories' is set to 'none', only a search summary is returned.
  * @method
  * @memberof queryBuilder#
  * @param {documents.categories}  [categories] - the categories of information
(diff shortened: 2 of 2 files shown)


kcoleman-marklogic:
I think what you have here is OK, but I wonder: 

Are you certain that it is necessary/correct to enumerate those specific options, or is it the case that using withOptions at all (regardless of the options you specify) causes the search response to be returned?

For example, the way to say "just return the search response and no documents" is to use `withOptions({category: 'none'})`.

I am not sure of the answer. I'm just wary of enumerating if it is not necessary.

wooldridge:
@kcoleman-marklogic, is there a good place in docs.marklogic.com to link to from the withOptions() description where a user can get more information about how all the properties (or maybe just the key ones) in that method work? To answer questions like, what are the available search options and how do they work? What do you think about doing that? I think Sudhakar's original bug submission points out that this area is complicated and a bit confusing to many who are just looking at the Node.js docs for withOptions().

wooldridge:
Thanks for your help, @kcoleman-marklogic. I did some more testing with withOptions(). As you point out, the  `{category: 'none'}` option gives you the search summary similar to what is seen with the other four options I mention. The options I don't mention return result docs (in some form or another) but no search summary. I want to address the original issue, which was confusion about when and why summaries appear. I will add a bit more about `{category: 'none'}`.

Issue #375, opened 2017-02-21
Review and address static analysis tool results


wooldridge:
We have a static analysis tool that analyzes source code and flags potential defects. It uses plug-ins such as JSHint as well as its own tests. We need to review the scan results of the Node.js Client API  source code and evaluate what fixes should be made. This is based on discussion with Tom T. on 2017-02-21.

wooldridge:
Need to confirm that issues addressed in #390 are no longer seen in static analysis report next time it runs.

wooldridge:
Currently seeing issues in results for the Node.js Client API:

https://bugtrack.marklogic.com/45957

wooldridge:
Ongoing task. Addressed issues for 2.1.1 timeframe, moving out to next release.

PR #390, opened 2017-03-12
fix(#375): remove or fix dead code uncovered by static analysis

wooldridge:


wooldridge:
Addresses some issues uncovered in #375

Issue #327, opened 2016-11-14
Release notes


jmakeig:
* High-level description
* Most important new/changed functionality
* Detailed change log
* Links to docs, tutorials, MLU, etc.

jmakeig:
@wooldridge Whose responsibility is this? Docs? QA? Dev?

wooldridge:
@jmakeig, I can take a stab at this. Will solicit input from @kcoleman-marklogic.

jmakeig:
It doesn't need a lot of details or “marketing”, just an accounting of what’s changed for developers. The [Data Hub Framework has a good example](https://github.com/marklogic/marklogic-data-hub/releases).

kcoleman-marklogic:
You can take a look at the README from the Java Client API, too. It hasn't been updated for this release yet, but the past releases are entirely suitable references.

jmakeig:
@wooldridge In the interest of operating more like other projects on GitHub, can we move the release-specific stuff out of the main README and put it into the release landing page? I wouldn’t expect the README to change much from release-to-release: 
- What is this thing? 
- How do I install it? 
- What’s the Day 0 “Hello, world!” interaction
- Where can I go for more information?
- Legal disclaimers

wooldridge:
I created a draft release page for the 2.0.0 release:

https://github.com/marklogic/node-client-api/releases

I added the compatibility info and new features there. Also included a link to the complete changelog, which will work after 2.0.0 is cut

I updated the README to remove the compatilibility info for 2.0.0 and the new features, which were moved to the release page: 

https://github.com/marklogic/node-client-api/blob/develop/README.md

Also added new items to the resources list (e.g., MLU resources) and made minor edits across the board.

kcoleman-marklogic:
Looks good to me, but I have one question:

Are we going to leave the EA releases on the Releases page? They seem more like internal detritus than anything else.

wooldridge:
@kcoleman-marklogic I agree about the EA releases cluttering the release page. Makes sense to me to remove EA releases from that page once the corresponding public release has come out. Users can still access EAs via tags if they need to.

jmakeig:
> Makes sense to me to remove EA releases from that page once the corresponding public release has come out.

@wooldridge, sounds prudent.

wooldridge:
Sounds like we're OK with these release notes as they are currently, will close issue.

PR #397, opened 2017-03-14
fix(#303): Update README with compatibility info

### LINKED ISSUE #303: Earlier versions of Client API should work with later versions of MarkLogic
Version 1.x of the Node.js Client API (compatible with ML8) should also work with ML9, but not support new features available in ML9. (If developers need to use new ML9 features, they will be required to upgrade to 2.x.)

This task is to check 1.x compatibility with ML9 upon the release of ML9 to make sure this is the case. Per discussion with @jmakeig.

wooldridge:
Fixes #303

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ from Node.js applications.
 
 ## Status
 
-Release 2.0.1 of the MarkLogic Node.js Client API, compatible with MarkLogic
-versions 8 and 9. To use the new features available in 2.0.1, you will need to
-run MarkLogic version 9.
+Release 2.0.1 of the MarkLogic Node.js Client API, tested to work with MarkLogic
+versions 8.0-6.3 and 9.0-1. To use the new features available in release 2.0.1,
+you will need to run MarkLogic version 9.
 
 ## New Features in Release 2.0.1
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,9 @@ from Node.js applications.
 
 ## Status
 
-Release 2.0.1 of the MarkLogic Node.js Client API
+Release 2.0.1 of the MarkLogic Node.js Client API, compatible with MarkLogic
+versions 8 and 9. To use the new features available in 2.0.1, you will need to
+run MarkLogic version 9.
 
 ## New Features in Release 2.0.1
 
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -131,6 +131,7 @@ function wipeOutputTransform(/*headers, data*/) {
  * {@link https://www.promisejs.org/|Promises} object for chaining
  * actions with then() functions.
  * @name ResultProvider#result
+ * @since 1.0
  * @function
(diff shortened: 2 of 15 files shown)


wooldridge:
@ayuwono, what's the easiest way to run Node.js Client API tests on the earliest and latest versions of ML8? I would like to substantiate that the upcoming release works on all versions of 8, and proving that it works on the earliest and latest would be strong proof of this. I've actually already run the tests on 8.0-6.3, so what I'm really looking for is the first version of ML8.

wooldridge:
@jmakeig, here's what we get in the current docs generation when we use  [@since](http://usejsdoc.org/tags-since.html)):

20170315_since_jsdoc

It would be great to add that info across the API docs. It would take some sleuthing to determine each method came into being.

kcoleman-marklogic:
At this stage, you can probably get away with tagging everything with 1.0, except what's new in 2.0-1. The whole API was new for 8.0-1, and while there were a few functional changes (like being able to pass multiple URIs to Documents.remove), I don't think any new functions or interfaces were added in 8.0-x.

The thing I always wonder about with a feature like this (we get requests for the XQuery docs, too) is what is the contract, really? Like in the example I gave above, Documents.remove was there from the beginning, but had its functionality extended. Does that get a version bump or not? Just one of those "inquiring minds want to know" things. :)

ayuwono:
@wooldridge, are you looking for the earliest version of ML8 (like 8.0-6.1)? You can find it at /project/engineering/builds/

wooldridge:
I've updated the JSDoc information with  [@since](http://usejsdoc.org/tags-since.html) tags. All methods and namespaces are marked version 1.0 except for the following:
```
documents.protect()
documents.wipe()
documents.protectResult()
documents.wipeResult()
patchBuilder.patchBuilderMetadataValues
patchBuilderMetadataValues.add()
patchBuilderMetadataValues.remove()
patchBuilderMetadataValues.replace()
queryBuilder.falseQuery()
queryBuilder.geospatialRegion()
queryBuilder.minDistance()
queryBuilder.trueQuery()
queryBuilder.MinDistanceParam
```

wooldridge:
I also updated the README to specify that 2.0.1 has been tested successfully with 8.0-6.3 and 9.0-1, which is true.  I discussed this with @ehennum, and he said this aligns with the rolling upgrade requirements for the client APIs: MarkLogic must offer clients that work with the latest release of the current version (e.g. 9.0-1) and the latest release of the previous version (e.g. 8.0-6.3). 

Erik and I also discussed the need for specifying version compatibility with each our tests for more efficient testing of API compatibility with MarkLogic versions. I could then run 2.0.1 on ML 8.0-6.3 and have it skip the tests for new features in 9.0-1.

wooldridge:
I'm going to merge this and close the associated issue (#303). We can fine tune the README as @jmakeig prepares release notes (https://github.com/marklogic/node-client-api/issues/327).

PR #399, opened 2017-04-24
Fix links to static methods in marklogic namespace

wooldridge:


wooldridge:
Fixes (part of) https://github.com/marklogic/node-client-api/issues/231

PR #415, opened 2017-07-27
Handle removing nonexistent document

### LINKED ISSUE #407: Implement check=exists param to be passed to delete requests to throw error
DELETE requests can now opt into a NOT FOUND error instead of idempotent silence when the document doesn't exist by passing a check=exists uri parameter

https://bugtrack.marklogic.com/44116

wooldridge:
The REST API now supports a "check=exists" parameter for document deletes.
Use this to recognize when a document being removed does not exist and return
"removed: false" in the response payload. Fixes #407

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -1381,13 +1381,15 @@ function removeOutputTransform(headers/*, data*/) {
   /*jshint validthis:true */
   var operation = this;
 
+  var removed = (operation.responseStatusCode === 204) ? true : false;
+
   if (operation.contentOnly === true) {
     return operation.uris;
   }
 
   var wrapper = {
     uris:    operation.uris,
-    removed: true
+    removed: removed
   };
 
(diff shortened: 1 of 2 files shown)


wooldridge:
@marklogic-builder, the implementation of this patch is ON HOLD. See #407.

jmakeig:
> Can one of the admins verify this patch?

What are you? Which planet has sent you? What is your intent? Do you come in peace?

kcoleman-marklogic:
Bad bot. Kill it! [Kill it with fire!](http://i0.kym-cdn.com/photos/images/original/000/128/623/kill-it-with-fire-demotivational-poster-1235695993.jpg)

ndw:
Nuke it from space. It's the only way to be sure.

jmakeig:
@ndw, are you talking about the nonexistent document or the alien bot?

Issue #421, opened 2017-09-25
Refactor operation.queueDocument()


wooldridge:
See: https://github.com/marklogic/node-client-api/blob/master/lib/operation.js#L93

This is part of a larger refactoring of the dispatch logic for object streaming in responder.js (which calls operation.queueDocument()). 

In operation.queueDocument(), separate queue management from object construction. This will allow the object stream dispatch to use object construction only. The promise stream dispatch will continue to use both queue management and object construction.

wooldridge:
Addressed by #422 and can be closed.

Issue #362, opened 2017-02-04
The example query-builder.js not completing


wooldridge:
When running all the examples at once, as described in the README, or when running query-builder.js alone, execution dies near the end of query-builder.js:
```
mwooldri$ node examples/query-builder.js 
Query documents with the Query Builder
Mayotte at /countries/mf.json
Seychelles at /countries/se.json
Morocco at /countries/mo.json
Burkina Faso at /countries/uv.json
Algeria at /countries/ag.json
Djibouti at /countries/dj.json
Cote d'Ivoire at /countries/iv.json
Mauritania at /countries/mr.json
Guinea at /countries/gv.json
Mali at /countries/ml.json
mwooldri$
```
It should end like this:
```
...
Mali at /countries/ml.json
Niger at /countries/ng.json
done
mwooldri$ 
```
All other examples complete successfully when run by themselves.

When query-builder.js is edited to use a promise instead of a stream (similar to how query-by-example.js uses a promise), the script completes successfully. So it appears to be an issue with streams in query-builder.js. 

I'm running Node.js v6.5.0 and npm v3.10.3.

wooldridge:
@ayuwono, with this fix we've refactored how we check if a multipart object stream is finished. Can you add the following test to check that this works with all cases?

Get results that have two documents with *metadata only* (no content) as an *object stream*. (This will test the situation when metadata is the last part in the multipart queue.)

Thanks to @ehennum for his help on this!

ayuwono:
Hi @wooldridge,

Is this fixed? I tried to run the example query-builder.js and it's still not completing

```
[ayuwono@ayuwono-z620 node-client-api]$ node examples/query-builder.js
Query documents with the Query Builder
Mayotte at /countries/mf.json
Seychelles at /countries/se.json
Morocco at /countries/mo.json
Burkina Faso at /countries/uv.json
Algeria at /countries/ag.json
Djibouti at /countries/dj.json
Cote d'Ivoire at /countries/iv.json
Mauritania at /countries/mr.json
Guinea at /countries/gv.json
Mali at /countries/ml.json
[ayuwono@ayuwono-z620 node-client-api]$
```

wooldridge:
Update, the query-builder.js example should run like this, returning the first 10 results, so ending at "Mali at /countries/ml.json" followed by "done":
```
MacPro-3170:node-client-api-examples mwooldri$ node examples/query-builder.js 
Query documents with the Query Builder
Mayotte at /countries/mf.json
[...]

PR #422, opened 2017-09-28
Refactor promise and object stream dispatches

wooldridge:
- Simplify the multipart object stream path. No longer use QueuedReader; instead write directly to the through2 output stream.
- Separate  object creation and object queuing for the multipart promise and multipart object stream cases. operation.makeObject() handles object creation; queuing (or writing) of the made objects is handled in the responder.
- Track metadata objects in a this.nextMetadataBuffer property in operation. 
- Add comments.

### DIFF
--- a/lib/responder.js
+++ b/lib/responder.js
@@ -412,8 +412,6 @@ MultipartDispatcher.prototype.chunkedStream = function dispatchMultipartChunkedS
       partReadStream.pipe(outputStream, {end: false});
     } else if (hasParsed) {
       responseFinisher();
-    } else if (!hasEnded) {
-      parser.emit('drain');
     }
   };
 
--- a/lib/responder.js
+++ b/lib/responder.js
@@ -462,6 +462,8 @@ MultipartDispatcher.prototype.objectStream = function dispatchMultipartObjectStr
   var hasParsed = false;
   var hasEnded  = false;
 
+  var isConcatenating = false;
+
   var responseFinisher = function objectResponseFinisher() {
 
     if (!partReaderQueue.hasItem() && hasParsed && hasEnded) {
@@ -511,10 +513,11 @@ MultipartDispatcher.prototype.objectStream = function dispatchMultipartObjectStr
     }
 
     partReaderQueue.removeFirst();
+    isConcatenating = false;
 
     // If item avail, concat-stream it with callback to finisher
     if (partReaderQueue.hasItem()) {
-
+      isConcatenating = true;
       var partRead = concatStream(partFinisher);
       partRead.on('error', errorListener);
       var partReadStream = partReaderQueue.getFirst();
@@ -540,6 +543,7 @@ MultipartDispatcher.prototype.objectStream = function dispatchMultipartObjectStr
     partReaderQueue.addLast(partReadStream);
 
     if (partReaderQueue.isLast()) {
+      isConcatenating = true;
       var partRead = concatStream(partFinisher);
       partRead.on('error', errorListener);
       partReadStream.pipe(partRead);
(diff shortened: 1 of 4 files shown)


lib/responder.js
```diff
@@ -248,17 +276,18 @@ MultipartDispatcher.prototype.promise = function dispatchMultipartPromise(
   var objectQueue     = new FifoQueue(2);
   var partReaderQueue = new FifoQueue(3);
 
-  var metadataBuffer  = null;
-
   var parsingParts = 0;
   var parsedParts  = 0;
 
   var hasParsed = false;
   var hasEnded  = false;
```
ehennum: Is there a need for hasEnded (and the end listener) in the Promise case?

lib/responder.js
```diff
@@ -418,27 +456,77 @@ MultipartDispatcher.prototype.objectStream = function dispatchMultipartObjectStr
...
+
+    var writeResult = null;
+    if (madeObject !== null && madeObject !== undefined) {
+      writeResult = operation.outputStream.write(madeObject);
+    }
+
+    // Manage backpressure
+    if (writeResult === false) {
+      // Only pause stream if not all parsed
+      if (hasParsed && !partReaderQueue.hasItem()) {
```
ehennum: Maybe test whether the HTTP ReadableStream has ended?  In which case there's nothing to pause?

lib/responder.js
wooldridge: We currently have a responseEndListener that sets hasEnded in the multipart promise case. Are saying when that happens, make sure to drain Dicer in case that stream is currently paused?

lib/responder.js
```diff
@@ -447,49 +535,31 @@ MultipartDispatcher.prototype.objectStream = function dispatchMultipartObjectStr
...
-      responseEndListener = null;
-      partTransform       = null;
-
-      operation.outputStream.end();
-    } else if (!hasEnded && parsedParts === parsingParts) {
-      parser.emit('drain');
-    }
-
+  var drainListener = function objectDrainListener() {
+    response.resume();
```
ehennum: Two belated questions

*  Should the drain listener resume only if the response has not ended? 
*  Should the drain listener start a concat stream on the next part?
*  To be more resillent on the previous bullet, should the dispatcher have an additional flag in its state to keep track of whether the first part is waiting or not?

lib/responder.js
ehennum: Sorry for the inclarity.  I'm saying that I believe there's never any point in firing the drain event -- not on Dicer or any other WritableStream.  My reading of https://nodejs.org/dist/latest-v8.x/docs/api/stream.html#stream_event_drain is that the drain event is a notification fired by the WriteableStream and not a request to the WriteableStream. In other words, the existing code was grossly mistaken in firing this event.

wooldridge: OK, that's how I understand it, too. I will remove the firing of the "drain" event in the chunked case, I'd overlooked that. And since we do a direct pipe of data in the chunked case, and pipe handles backpressure, we should be OK there.

lib/responder.js
wooldridge: Thanks so much for the comments. I addressed the first two bullets with https://github.com/marklogic/node-client-api/pull/422/commits/fdde417708efa010a9c7c5e37daa74acd69aad3d. For the third bullet, seems like existing queue methods are enough to keep track. By "first part," do you mean the first part initially added to the queue or just a part in the queue during a resume step?

ehennum: My concern is whether the drain event might ever be fired when the first part is already concatenating. In principle, that shouldn't happen, but sometimes it's useful to guard against things that shouldn't happen.  If an existing flag already records whether the first part is concatenating, then nothing else is needed.

wooldridge: Does setting an isConcatenating flag here address your concerns? https://github.com/marklogic/node-client-api/pull/422/commits/ea415d597944b097123d55a015e9aa1fed5a1834

ehennum: Most definitely.  Thanks.

wooldridge:
Fixes #362 
Fixes #421

Issue #296, opened 2016-09-15
Add support for point-in-time queries


wooldridge:

wooldridge:
@ayuwono, see the functional spec on the wiki for details about the methods updated.

Relevant commits: 
f7b83e7962e9a6108bfaa4fe4085af70ef553fc4
f7f8d12c218006d99f82e3f7f2a25855dc4523eb
fd60f29c99590f41ac10546e42941726ca42b355

ayuwono:
Added tests on test-complete/nodejs-pitq-*.js

PR #433, opened 2017-10-23
Add support for point-in-time queries using timestamp

wooldridge:
The following methods are affected:
documents#query
documents#read
graphs#list
graphs#read
graphs#sparql
values#read
(rows#query is also affected but will be dealt with in Optic branch)

Includes adding big-integer dependency for manipulating numbers larger than Node.js's max allowable value
Needed for testing timestamp


wooldridge:
@ehennum registered his approval via email.

Fixes #296

Issue #284, opened 2016-08-24
Inconsistent count for batch load


tpiros:
Hi,

Give the following sample code taken from [the examples](https://github.com/marklogic/node-client-api/blob/8f63f6b531258c733be3ff0e9a992fcd7d444f37/examples/before-load.js) I have experienced the following behaviour.

I have updated the above code in the following way:

``` javascript
// I have added a counter -->
var batchSize = 100;
var counter = 0; //added a counter
// I then increase the counter -->
db.documents.write(buffer).result(function(response) {
        console.log(
            'done loading:\n'+
            response.documents.map(function(document) {
              counter++; //increasing the counter here
              return document.uri;
              }).join(', ')+'\n'
            );
        writeBatch(filenames, i + 1);
// and finally print out some debug info:
function writeBatch(filenames, batchFirst) {
  if (batchFirst >= filenames.length) {
    console.log('filenames.length ==> ', filenames.length); //print filenames.length
    console.log('counter ==> ', counter); //print counter
    console.log('done loading example data to');
```

Interestingly each time I execute the code I get different results:

``` bash
$ node before-load.js
filenames.length ==>  261
counter ==>  259
$ node before-load.js
filenames.length ==>  261
counter ==>  261
$ node before-load.js
filenames.length ==>  261
counter ==>  260
```

As you can see the counter gives a different value after each execution.

To debug this further I have reduced the number of documents in the data folder to have only 21 documents and reduced the batch size to be 10. This is what has happened:

``` bash
$ node before-load.js
loading example data
loading batch from /countries/aa.json to ao.json
done loading:
/countries/aa.json, /countries/ae.json, /countries/ac.json, /countries/af.json, /countries/ag.json, /countries/aj.json, /countries/al.json, /countries/am.json, /countries/an.json, /countries/ao.json

loading batch from /countries/as.json to bb.json
done loading:
/countries/as.json, /countries/ar.json, /countries/at.json, /countries/aq.json, /countries/au.json, /countries/av.json, /countries/ax.json, /countries/ay.json, /countries/bb.json

loading batch from /countries/bc.json to bc.json
done loading:
/countries/bc.json

filenames.length ==>  21
counter ==>  20
done loading example data
```

The file `ba.json` is missing - and it hasn't been loaded.

Second time executing the same command:

``` bash
$ node before-load.json
loading example data to
loading batch from /countries/aa.json to ao.json
done loading:
[...]

PR #448, opened 2017-10-31
Fix async loading of example files

wooldridge:
Use Promise.all() to write batches of files only after all example files in batch have been read.


wooldridge:
Fixes #284

PR #471, opened 2017-12-01
Add TDE test and rows example

wooldridge:


ehennum:
That looks reasonable to me.  The plan-views.js unit tests already test TDE templates via the row manager, but having the row manager test cover the most important case (views) is a good idea so that's testable in isolation.

Thanks for enhancing the examples to demonstrate Optic.

PR #482, opened 2017-12-08
Add database param for rows queries if exists

### LINKED ISSUE #478: Rows queries not supporting database defined in client connection
The /v1/rows REST endpoint supports a database parameter for specifying an alternative db, but the Node.js Client API currently does not pass along the db from the DatabaseClient connection configuration. It should.

To reproduce, just make an Optic query using a db client whose connection config specifies an alternative db. For example, set up the SQLdata db and content from the SQL quick start (http://docs.marklogic.com/guide/sql/setup) and then try to query it via port 8000 using Node.js. 

I used that setup and this query:
```
const marklogic = require('marklogic');
const my = require('./my-connection.js');

const db = marklogic.createDatabaseClient(my.connInfo);
const pb = marklogic.planBuilder;

db.rows.query(
  pb.fromView('main', 'employees')
    .select(['EmployeeID', 'FirstName', 'LastName'])
    .orderBy('EmployeeID')
    .limit(3)
).then( function(rows) {
    console.log(JSON.stringify(rows, null, 2));
}).catch(
  function(error) {
    console.log('ERROR:');
    console.log(JSON.stringify(error, null, 2));
});
```

Originally, my db config looked like this:
```
module.exports = {
  connInfo: {
    host: 'myhost',
    port: 8000,
    user: 'myuser',
    password: 'mypassword',
    database: 'SQLdata'
  }
};
```
If you try that, the server squawks that there is no suitable view because it is trying to use Documents. It only started working when I created a REST instance with the following config and changed my connection info to match in Node.js.
```
{"rest-api": {
  "name": "optic-test",
  "database": "SQLdata",
  "port": "8005"
}}
```
I had also confirmed that a raw rest call that included the equivalent plan and `database=SQLdata` worked fine.

(Sorry, @wooldridge , I had to tag this as 2.1.1 because there's not 2.1.2 milestone defined yet.)

wooldridge:
Also add test for this. Fixes #478.

### DIFF
--- a/lib/rows.js
+++ b/lib/rows.js
@@ -202,7 +202,8 @@ function queryRowsImpl(builtPlan, streamType, options) {
     }
   }
 
-  requestOptions.path = endpoint;
+  // Specify database if defined in client connection
+  requestOptions.path = mlutil.databaseParam(connectionParams, endpoint, '&');
 
   const operation = new Operation(
     'query rows', this.client, requestOptions, 'single', responseType
--- a/test-basic/client.js
+++ b/test-basic/client.js
@@ -23,10 +23,11 @@ var q = marklogic.queryBuilder;
 
 var YAgent = require('yakaa');
 
-// TODO: setup should create user with eval-in role
+// 'rest-evaluator' user can evaluate against an alternate db:
+// http://docs.marklogic.com/guide/rest-dev/intro#id_72318
 var connection = {
-    user:     'admin',
-    password: 'admin'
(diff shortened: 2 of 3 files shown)


wooldridge:
@ehennum, the update to have the client db tests use the 'rest-evaluator' user passes those tests.

ehennum:
Thanks for fixing up the unit test.

ayuwono:
Ok, looks good

wooldridge:
Approved by mrtriage.

Issue #459, opened 2017-11-14
Increase the supported Node.js version to 6.9.0


wooldridge:
This will ensure ES6 Symbol.hasInstance support, which is required for the Optic implementation in Node.js.

Run appropriate tests and ensure that the new version combinations (MarkLogic and Node.js) meet rolling upgrade requirements.

wooldridge:
Updating and testing the supported Node.js version should come after #460 (update dependencies) and #461 (update shrinkwrap).

The package.json will also need updating, currently:
```
"engines": {
    "node": ">=6.3.1"
  }
```

wooldridge:
According to @ayuwono on 2017-11-27:

"The QA regression has been running with node 6.9.0 since last week. It's been stable so far."

Still need to confirm that we meet the version requirements for rolling upgrade. Dependencies (#460) and shrinkwrap (#461) have been updated and pass tests.

PR #491, opened 2018-01-18
Up Node.js version support to 6.9.0.

wooldridge:


wooldridge:
See: https://github.com/marklogic/node-client-api/issues/459

PR #515, opened 2019-01-21
Fixes #514: Adds missing deps and npm scripts

### LINKED ISSUE #514: Missing dependencies, unnecessary global gulp
On a fresh install, there seem to be two missing dependencies for successfully generating the documentation:

* `moment` (latest is 2.23.0)
* `sanitize-html`(latest is 1.20.0)

Also, the requirement for having `gulp` installed globally is unnecessary, and should be wrapped in some entries in the `scripts` block in `package.json`, like so:

```
{
  "scripts": {
    "doc": "gulp doc",
    "test:setup": "node etc/test-setup.js",
    "test:teardown": "node etc/test-teardown.js",
    "test": "gulp test"
}
```

pete-otaqui:
Added the missing `moment` and `sanitize-html` dependencies.

Wrapped the various script executables into the `package.json` scripts
block, also updating the README.md to reflect it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,11 @@ in the online resources. To run the examples, follow the instructions here:
 
 ### Generating Documentation Locally
 
-After installing the project dependencies (including the [gulp](http://gulpjs.com/) 
-build system), you can build the reference documentation locally from the root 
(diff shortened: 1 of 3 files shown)


ehennum:
Should only affect documentation generation and should cause no issues there.

Issue #508, opened 2018-10-03
add beforeQuery() and afterQuery() query constructors to queryBuilder


ehennum:
The signatures might be similar to:

    beforeQuery(timestamp) → {queryBuilder.Query}
    afterQuery(timestamp) → {queryBuilder.Query}

where timestamp is an unsigned long stated as a string, emitting 

    {"before-query":{"timestamp":"..."}}
    {"after-query":{"timestamp":"..."}}

respectively.

See server bug 50902 for the details.

kcoleman-marklogic:
When this work gets done, please don't forget to notify TechPubs. Probably Eleanor Tesoro, but if in doubt, let Jane M. know.

georgeajit:
Tests are running fine.

PR #525, opened 2019-07-03
added beforeQuery and afterQueryquery to queryBuilder

anu3990:
Fix for https://github.com/marklogic/node-client-api/issues/508

### DIFF
--- a/lib/query-builder.js
+++ b/lib/query-builder.js
@@ -3140,27 +3140,21 @@ function TermDef(textQuery) {
 }
 util.inherits(TermDef, QueryDef);
 
+/**
+ * Builds a query to return fragments committed before a specified time.
+ * @method
+ * @memberof queryBuilder#
+ * @param {string, number or date} - referring to the time to find fragments committed before this
+ * @returns {queryBuilder.Query} a composable query
+ */
 function before() {
    var args = mlutil.asArray.apply(null, arguments);
    if(args.length === 0 || args[0] === null)
      throw new Error('missing Timestamp');
    if(args.length>1)
      throw new Error('Only one timestamp is allowed');
-   var arg = new String(args[0]);
 
-   if(!(typeof(args[0]) == 'string')) {
-     if(typeof(args[0]) == 'number') {
-      arg = new String(args[0]);
-     }
-     else if(args[0] instanceof Date) {
-      var epochMilli = args[0].getTime();
-      arg = new String(epochMilli);
-     }
-     else {
-      throw new Error('The timestamp argument needs to be either String, Number or Date');
-     }
-   }
-  return new BeforeQueryDef(new TimestampDef(arg));
+  return new BeforeQueryDef(new TimestampDef(args[0]));
 }
 /** @ignore */
 function BeforeQueryDef(timestampQuery) {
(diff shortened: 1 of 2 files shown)


lib/query-builder.js
```diff
@@ -3139,6 +3139,80 @@ function TermDef(textQuery) {
...
+/** @ignore */
+function BeforeQueryDef(timestampQuery) {
+  if (!(this instanceof BeforeQueryDef)) {
+    return new BeforeQueryDef(timestampQuery);
+  }
+
+  if (timestampQuery instanceof TimestampDef) {
+    this['before-query'] = timestampQuery;
+  } else {
+    throw new Error('missing query');
```
ehennum: Would the error be more easily understood as 'missing timestamp value for query' both here and in AfterQuery?

lib/query-builder.js
```diff
@@ -3139,6 +3139,80 @@ function TermDef(textQuery) {
...
 }
 util.inherits(TermDef, QueryDef);
+
+function before() {
+   var args = mlutil.asArray.apply(null, arguments);
+   if(args.length === 0 || args[0] === null)
+     throw new Error('missing Timestamp');
+   if(args.length>1)
+     throw new Error('Only one timestamp is allowed');
+   var arg = new String(args[0]);
```
ehennum: Would it make sense to pass arg[0] without wrapping in new String() to the TimestampDef constructor and move the arg testing and conversion into the TimestampDef constructor so we maintain that code in one spot and so the TimestampDef object is fully responsible for its data?

Should the TimestampDef constructor allow for either a string primitive or a String object without conversion (in particular, without constructing a new String() object)?

A string primitive and a String object are different types in JavaScript.  For an example, the following code snippet:

```
const primitive = "value";
const object    = new String(primitive);
[typeof primitive, primitive instanceof String, typeof object, object instanceof String];
```

yields `["string", false, "object", true]`

If so, the TimestampDef constructor would skip conversion for if arg instanceof String || typeof arg == 'string'

test-basic/query-builder.js
```diff
@@ -2001,7 +2001,50 @@ describe('query-builder', function() {
         {'false-query':null}
         );
   });
-
+  it('should create a before-query', function(){
+   var d = new String('foo');
```
ehennum: Could the test value be something like '1562178870875' instead of 'foo'?

We don't currently check that the value is an unsigned long but might someday and, even if we don't, it's easier for someone else to understand the test if the value is valid for the data type.

test-basic/query-builder.js
```diff
@@ -2001,7 +2001,50 @@ describe('query-builder', function() {
...
+   assert.deepEqual(
+        q.before('foo'),
+        {'before-query':{timestamp:d}}, 'Error in Before-Query with String input'
+        );
+  });
+  it('should create a before-query', function(){
+   var d = new String(1562178870875);
```
ehennum: This test value is a string.  

A number would be either or (better) both of `var d = 1562178870875;` and `var d = new Number(1562178870875);`

test-basic/query-builder.js
```diff
@@ -2001,7 +2001,50 @@ describe('query-builder', function() {
...
+   assert.deepEqual(
+        q.before(1562178870875),
+        {'before-query':{timestamp:d}}, 'Error in Before-Query with Number'
+        );
+  });
+  it('should create a before-query', function(){
+   var input = new Date(2019, 7, 3);
+   var d = new String(1564815600000);
```
ehennum: Would the test be easier to understand as `var d = new String(input.getTime());`?

test-basic/query-builder.js
```diff
@@ -2001,7 +2001,50 @@ describe('query-builder', function() {
...
+   assert.deepEqual(
+        q.before(input),
+        {'before-query':{timestamp:d}}, 'Error in Before-Query with Date input'
+        );
+  });
+  it('should create an after-query', function(){
```
ehennum: The feedback on the before query tests also applies to the after query tests

Issue #499, opened 2018-06-20
Allow passing transaction object between services


g0liushi:
When (micro) services are passing transaction objects via http,  the transaction object became a plain JS object (loosing the Transaction class),  and will cause an error when used by the downstream service: `Error: can only undefined transaction with string or Transaction object identifier: object` (because `getTxidParam` in `lib/mlutil.js ` checks the `instanceof Transaction`)

My current solution is either just use the transaction string, or import `lib/mlutil.js` and initiate a new Transaction class locally. 

Is there a better way to handle this? Or is it possible to allow passing a plain JS object and do more sanity checks in `getTxidParam` etc.?

ehennum:
I'm wondering whether the contexts that take a transaction object might do the following:

*  Sniff the type of the transaction object (as currently)
*  If the object doesn't have the transaction type, fall back to passing the object to the transaction object constructor

Possibly this checking logic might be put into a checker() function in the library with the transaction object constructor.

georgeajit:
Works fine.

PR #529, opened 2019-07-29
Allowing to pass transaction object between services.

anu3990:
Fix for - https://github.com/marklogic/node-client-api/issues/499

### DIFF
--- a/test-basic/transactions.js
+++ b/test-basic/transactions.js
@@ -232,7 +232,6 @@ describe('transaction', function(){
     it('should read the same file as written when using mlutil.convertTransaction', function(done){
       var txn = null;
       var tid = null;
-      var cookies = null;
       db.transactions.open(true)
       .result(function(response) {
         txn = response;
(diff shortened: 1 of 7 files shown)


test-basic/transactions.js
```diff
@@ -226,5 +228,43 @@ describe('transaction', function(){
...
+        return db.documents.write({
+          txid: txn,
+          uri: uri,
+          contentType: 'application/json',
+          content: {txKey: txn.txid}
+          }).result();
+        })
+      .then(function(response) {
+        tid = txn.txid;
+        cookies = txn.cookies[0];
```
ehennum: Would it be a better test to construct a new JavaScript literal here?  As in:

```
var  txnRaw = {txid: txn.txid, cookies: txtn.cookies};
var convertedTransaction = mlutil.convertTransaction(txnRaw);
```

test-basic/transactions.js
```diff
@@ -226,5 +228,43 @@ describe('transaction', function(){
...
+        return db.transactions.read(convertedTransaction).result() ;
+        })
+      .then(function(document) {
+        var status = document['transaction-status'];
+        var transactionId = status['transaction-id'];
+        var host = status['host'];
+        var hostId = 'HostId='+host['host-id'];
+
+        tid.should.equal(transactionId);
+        cookies.should.equal(hostId);
```
ehennum: Given that the sequence of cookies is under the control of the load balancer, it might be sufficient to skip the cookie inspection and just verify that the read request in the transaction succeeded and returned the same transaction id.

anu3990:
Thanks Erik.! 
Merging the pull request after adding the suggestions.

Issue #341, opened 2017-01-07
queryBuilder.extract cannot accept namespace bindings


kcoleman-marklogic:
`queryBuilder.extract` thinks you can feed it namespace bindings for use in the extract path expressions. The original spec, the impl, and the JSDoc all assert this. Three cheers for consistency and ease of use.  Alas, this does not, in fact, work. 

When I tried this, the node client returned a 500 (how rude) and the server log contained an XDMP-UNBPRFX error. If I pre-define my namespace prefix on my group using the Admin UI, then the extract worked fine.

This behavior is consistent with the `extract-document-data` option documentation, which says you have to pre-define namespace bindings when using JSON options. The JSON option syntax does not allow for specification of namespace bindings. See the following:

http://docs.marklogic.com/guide/search-dev/appendixa#id_98906

I will leave it to you, Erik & Co. to ponder whether or not we want to address the underlying systemic problem. If we do not make a systemic change, you need to change the impl and the JSDoc for `queryBuilder.extract`, and I need to change its documentation in the Node.js dev guide.

Below is the example code I was trying to run. For purposes of this problem, the content you're querying is irrelevant.
```javascript
const marklogic = require('marklogic');

// MODIFY THIS VAR TO MATCH YOUR ENV
const connInfo = {
    host: 'localhost',
    port: 8000,
    user: 'admin',
    password: 'admin',
    database: 'Documents'
  };
const db = marklogic.createDatabaseClient(connInfo);
const qb = marklogic.queryBuilder;

db.documents.query(
  qb.where(
    qb.collection('person-envelopes'),
    qb.parsedFrom('entity-type:person AND washington',
      qb.parseBindings(
        qb.value(
          qb.element(
            qb.qname('http://marklogic.com/entity-services','title')),
          qb.bind('entity-type'))))
  ).slice(qb.extract({
       paths: ['//es:instance/(Person)'],
       selected: 'include',
       namespaces: {'es': 'http://marklogic.com/entity-services'}
     }))
).result( function(results) {
  for (let result of results) {
    console.log(JSON.stringify(result));
  }
});
```

kcoleman-marklogic:
Just wanted to add one more thing about this bug: The Node.js API does not, AFAIK, expose the /v1/config/namespaces endpoint, so you cannot workaround this problem wholly within the purview of Node.js. 

[...]

PR #531, opened 2019-08-09
queryBuilder.extract now accepts namespace bindings

anu3990:
Fix for - https://github.com/marklogic/node-client-api/issues/341

### DIFF
--- a/test-basic/documents-query.js
+++ b/test-basic/documents-query.js
@@ -115,6 +115,11 @@ describe('document query', function(){
           collections: ['extraCollection'],
           contentType: 'application/xml',
           content:     '<extraRoot><extraMatch>matchable extra</extraMatch></extraRoot>'
+        }, {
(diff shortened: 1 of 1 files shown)


test-basic/documents-query.js
```diff
@@ -497,6 +502,24 @@ describe('document query', function(){
...
+            q.word('target','match')
+          ).
+        slice(0, 1, q.extract({
+          selected:'include',
+          paths:'//abc:elem',
+          namespaces: {'abc': 'http://marklogic.com/test/abc'}
+          }))
+        )
+      .result(function(response) {
+        response.length.should.equal(1);
```
ehennum: Similar to the preceding extract test, should the test verify that document.content contains the expected string for the extracted XML?

anu3990: Yes that’s right. Adding the code for content check.

Issue #346, opened 2017-01-20
Test connection


jmakeig:
As a developer I want to debug connection issues. I don't want to have to do any real work (i.e. run a dummy query or test the existence of a document) because that's extra work and potentially introduces other errors. I just want to see if my auth*n and network are configured correctly. 

As an ETL tool maker, I want to provide a "Test Connection" button in my UI.

The Client REST API has the undocumented (?) `/v1/ping` endpoint that might be related.

Related to marklogic/java-client-api#636.

georgeajit:
Test runs fine.

PR #532, opened 2019-08-20
Adding feature for - "Test Connection"

anu3990:
Enhancement feature referring to - https://github.com/marklogic/node-client-api/issues/346

### DIFF
--- a/lib/marklogic.js
+++ b/lib/marklogic.js
@@ -376,7 +376,8 @@ function MarkLogicClient(connectionParams) {
  * Tests if a connection is successful.
  * @method DatabaseClient#testConnection
  * @since 2.1
- * @returns true if the connection is successful and false otherwise with an error message
+ * @returns an object with a connected property of true or false. In the false case 
+ *  it contains httpStatusCode and httpStatusMessage properties identifying the failure.
  */
 MarkLogicClient.prototype.testConnection = function testConnection() {
   var connectionParams = this.connectionParams;
@@ -384,7 +385,7 @@ MarkLogicClient.prototype.testConnection = function testConnection() {
   requestOptions.method = 'HEAD';
   requestOptions.path = mlutil.databaseParam(connectionParams, '/v1/ping', '?');
   var operation = new Operation(
-      'test operation', this, requestOptions, 'empty', 'single'
+      'test operation', this, requestOptions, 'empty', 'empty'
       );
   operation.statusCodeValidator = function testStatusCodes(statusCode, response){
     if (statusCode >= 300)
@@ -398,11 +399,10 @@ MarkLogicClient.prototype.testConnection = function testConnection() {
     } else {
        content = {connected: false,
        httpStatusCode:  this.responseStatusCode,
-       httpStatusMessage: operation.responseStatusMessage
-       }
-    var wrapper = {content: [content]};
+       httpStatusMessage: this.responseStatusMessage
+       };
     }
-   return wrapper;
+   return content;
   }
 return requester.startRequest(operation);
 };
--- a/lib/marklogic.js
+++ b/lib/marklogic.js
@@ -37,6 +37,8 @@ var queryBuilder         = require('./query-builder.js');
 var patchBuilder         = require('./patch-builder.js');
 var valuesBuilder        = require('./values-builder.js');
 var planBuilder          = require('./plan-builder.js');
+var Operation            = require('./operation.js');
+var requester            = require('./requester.js');
 
 /**
  * Provides functions to connect to a MarkLogic database and to build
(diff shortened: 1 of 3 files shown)


lib/marklogic.js
```diff
@@ -369,6 +371,42 @@ function MarkLogicClient(connectionParams) {
 
   this.logger = null;
 }
+
+/**
+ * Tests if a connection is successful.
+ * @method DatabaseClient#testConnection
+ * @since 2.1
+ * @returns true if the connection is successful and false otherwise with an error message
```
ehennum: Would it be more accurate to say that it returns an object with a connected property of true or false and, in the false case, httpStatusCode and httpStatusMessage properties identifying the failure?

lib/marklogic.js
```diff
@@ -369,6 +371,42 @@ function MarkLogicClient(connectionParams) {
...
+ */
+MarkLogicClient.prototype.testConnection = function testConnection() {
+  var connectionParams = this.connectionParams;
+  var requestOptions = mlutil.copyProperties(connectionParams);
+  requestOptions.method = 'HEAD';
+  requestOptions.path = mlutil.databaseParam(connectionParams, '/v1/ping', '?');
+  var operation = new Operation(
+      'test operation', this, requestOptions, 'empty', 'single'
```
ehennum: Should the last parameter be empty?  After experimenting with curl, it looks like the HEAD request doesn't return a body in either the success or failure cases.

lib/marklogic.js
```diff
@@ -369,6 +371,42 @@ function MarkLogicClient(connectionParams) {
...
+    return null;
+   };
+  operation.outputTransform = function addOutputTransform() {
+    var content = null;
+    if(this.responseStatusCode <300) {
+       content = {connected: true};
+    } else {
+       content = {connected: false,
+       httpStatusCode:  this.responseStatusCode,
+       httpStatusMessage: operation.responseStatusMessage
```
ehennum: Should the assignment use this.responseStatusMessage instead of operation.responseStatusMessage for consistency and to avoid side effects if the code is refactored to move the addOutputTransform() function out of the the testConnection() function in the future?

The this object refers to the object on which the function is called while operation is a closure on the local variable of testConnection()

lib/marklogic.js
```diff
@@ -369,6 +371,42 @@ function MarkLogicClient(connectionParams) {
...
+       }
```
ehennum: Should the previous line end with a semicolon to reinforce that it closes the object literal assigment starting on line 399?

lib/marklogic.js
```diff
@@ -369,6 +371,42 @@ function MarkLogicClient(connectionParams) {
...
+    var wrapper = {content: [content]};
+    }
+   return wrapper;
```
ehennum: I've two concerns:

1. The wrapper variable is assigned only inside the else branch but returned from ddOutputTransform() in the if case as well.  Wouldn't the return value be undef in that case?
2.  What is the rationale for wrapping the content value instead of just returning the content value?

test-basic/documents-quick.js
```diff
@@ -167,3 +167,67 @@ describe('quick path', function(){
...
+
+describe('test connection', function(){
+ var assert = require('assert');
+    it('should test the connection', function(done){
+     	var flag = false;
+     db.testConnection()
+      .result(function(response) {
+        flag = true;
+     }).then(function(){
+        assert(flag == true);
```
ehennum: Because the result() callback doesn't return a promise, I believe the then() callback will never execute.

Same comment on subsequent tests.

test-basic/documents-quick.js
```diff
@@ -167,3 +167,67 @@ describe('quick path', function(){
...
   });
 });
+
+describe('test connection', function(){
+ var assert = require('assert');
+    it('should test the connection', function(done){
+     	var flag = false;
+     db.testConnection()
+      .result(function(response) {
+        flag = true;
```
ehennum: Now that test connection returns an object in all cases, would a better test assert that response.connected is true?

test-basic/documents-quick.js
```diff
@@ -167,3 +167,67 @@ describe('quick path', function(){
...
+    var flag = false; 
+    var content = {connected: false,
+       httpStatusCode:  403,
+       httpStatusMessage: 'Forbidden'
+       };
+    var respContent = null;
+     db1.testConnection()
+      .result(function(response) {
+        flag = true;
+        respContent = response.content[0];
```
ehennum: Instead of having to get a content subproperty, would it be better to have the properties directly on the response. 

Same comment on the subsequent test.

anu3990:
Thanks Erik. ! 
Changed "test connection" to "check connection"

tdiepenbrock:
The documentation on this still refers to "testConnection".

jmakeig:
Ping @ehennum.

ehennum:
Please see Tom's feedback...

PR #533, opened 2019-09-03
Created a user and role to verify the connection under different

anu3990:
scenarios.

### DIFF
--- a/etc/test-config.js
+++ b/etc/test-config.js
@@ -38,6 +38,9 @@ var restTemporalPassword = 'x';
 
 var testServerName = 'unittest-nodeapi';
 
+var testUser = 'test-user';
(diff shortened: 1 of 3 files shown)


etc/test-config.js
```diff
@@ -94,5 +97,14 @@ module.exports = {
...
         ssl:      true
+    },
+testConnection: {
+	    host:     testHost,
+        port:     restPort,
+        user:     testUser,
+        password: testPassword,
+        authType: restAuthType,
+        rejectUnauthorized: false,
+        ssl:      true
```
ehennum: Should it be possible to use this test without an SSL connection?

anu3990: Yes. The tests are running fine without ssl field.

PR #546, opened 2020-04-10
Add semicolons for unterminated statements 

hansenmc:
instead of relying on automatic semicolon insertion, which can result in errors and unexpected behaviors.

### DIFF
--- a/config-optic/connectdef.js
+++ b/config-optic/connectdef.js
@@ -23,7 +23,7 @@ var plan = {
   user:     'admin',
   password: 'admin',
   authType: 'DIGEST'
-}
+};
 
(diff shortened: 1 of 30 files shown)


jmakeig:
Yes, explicit semis are best practice, but I’m not sure we’d want to go back and reformat source code just to add them. Did you uncover a bug caused by an ambiguity?

hansenmc:
No, linter just picked it up and it seems that the codebase generally does use `;`, so thought it would be best to be consistent.

Feel free to reject and throw away the changes if you don't think it's necessary

ehennum:
should cause no regressions

Issue #559, opened 2020-11-12
throw error on requests with a released client


jkerr5:
Handle "ECONNRESET" errors that can happen if a connection is suddenly closed out from the client. Provide a more descriptive error message indicating what happened. Here is an example of what gets reported now:

```
{ Error: socket hang upat createHangUpError (_http_client.js:254:15)at TLSSocket.socketCloseListener (_http_client.js:286:23)at emitOne (events.js:101:20)at TLSSocket.emit (events.js:188:7)at _handle.close (net.js:497:12)at TCP.done [as _onclose] (_tls_wrap.js:332:7) code: 'ECONNRESET' }
```

ehennum:
After a client is released, the client should throw an error on any call that attempts to use the client.

Given that the ECONNRESET errors thrown by Node.js may have many causes, it would risk providing bad guidance to make assumptions about the root cause of the error.

ehennum:
One possible fix would be to 

- add a getter for client.connectionParams that throws the friendly error if connectionParams is null
- refactor all code that reads client.connectionParams currently to call the getter instead
- modify releaseClient() to set connectionParams to null

PR #563, opened 2021-03-17
#559- throw error on requests with a released client

anu3990:


ehennum (approved):
Thanks for squaring away the fix

Issue #562, opened 2021-02-02
cts.query builder


ehennum:
Similar to the Java API, reuse the cts.query builder included with the Optic builder for search requests.

ehennum:
The Java equivalent:

https://github.com/marklogic/java-client-api/issues/1260

The solution might

- Add a ctsQueryBuilder() factory method to the marklogic.js exports
- From the factory method, return an object with the CtsExpr, FnExpr, GeoExpr, JsonExpr, MapExpr, MathExpr, RdfExpr, SemExpr, SpellExpr, SqlExpr, XdmpExpr, and XsExpr exports from plan-generated.js 
- In the Documents.prototype.query() method, accept a types.CtsQuery object as a built query and serialize it by applying JSON.stringify() to the return value of exportArg() from plan-builder-base.js in much the same way as Rows.prototype.query() but wrapping it with a "ctsast" JSON property similar to the Java implementation

PR #564, opened 2021-03-18
#562 - cts.query builder

anu3990:
cts.query builder

### DIFF
--- a/lib/ctsquery-builder.js
+++ b/lib/ctsquery-builder.js
@@ -18,7 +18,7 @@
 const bldrgen  = require('./plan-builder-generated.js');
 
     /**
-     * A helper for building the definition of a rows query. The helper is
+     * A helper for building the definition of a query. The helper is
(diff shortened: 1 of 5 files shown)


lib/ctsquery-builder.js
```diff
@@ -0,0 +1,120 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const bldrgen  = require('./plan-builder-generated.js');
+
+    /**
+     * A helper for building the definition of a rows query. The helper is
```
ehennum, line 21: In this context, it would be better to delete the word "rows" because we're getting documents instead of rows

ehennum (approved):
One trivial documentation change

lib/ctsquery-builder.js
anu3990: Thanks Erik.! Included the change.

Issue #565, opened 2021-03-22
Upgrade node-client-api to use node 14


anu3990:
Currently, the node-client-api uses node version -10. We are introducing compatibility with node version -14 to make use of new node features.

ehennum:
Here's the Node.js timeline:

https://nodejs.org/en/about/releases/

georgeajit:
Verified on 06/22/21 : https://jenkins.marklogic.com/view/Node%20API/job/NodeAPIUnitTests_develop_b10_0/3467/console

```
+ which node
/bin/node
+ node --version
v14.16.1
+ node etc/test-setup.js -u admin:admin
checking for REST users
```

PR #566, opened 2021-04-05
#565 - Upgrade node-client-api to use node 14

anu3990:


package.json
```diff
@@ -27,44 +27,45 @@
   ],
   "dependencies": {
     "concat-stream": "^2.0.0",
-    "deepcopy": "^2.0.0",
+    "deepcopy": "^2.1.0",
     "dicer": "^0.3.0",
-    "json-text-sequence": "^0.1.1",
+    "gulp-update": "0.0.2",
```
ehennum, line 32: Does the Node.js API have a direct dependency on gulp-update?  Or only an indirect dependency via gulp.

If the Node.js API doesn't make direct use of gulp-update, it's better not to list the dependency because requiring the correct version of gulp will also pick up the correct version of gulp-update.

package.json
```diff
@@ -27,44 +27,45 @@
   ],
   "dependencies": {
```
ehennum, line 28: If I recall correctly, this change to package.json updates all libraries with more recent versions. To put it the other way, libraries that aren't upgraded such as concat-stream are already the most recent available. I'm noting it to confirm that point because it is best to update libraries when possible to get the benefit of bug fixes in those libraries.

ehennum (approved):
Thanks for sorting through the dependencies.

package.json
anu3990: Done.

Issue #543, opened 2020-04-02
prefer ipv4 to ipv6 addresses


ehennum:
MarkLogic server prefers ipv4 addresses to ipv6 addresses.

Because the client's address factors into validation of the nonce in DIGEST authentication, the client must continue to use the same address while using the nonce.

In Java, it became necessary under heavy concurrency with DIGEST authentication to prefer ipv4 addresses explicitly on the client.

If that becomes necessary in Node.js, the HTTP request() in version 12 supports a lookup function:

https://nodejs.org/docs/latest-v12.x/api/http.html#http_http_request_options_callback

Presumably, that function has the same signature as dns.lookup() and could wrap a call to dns.lookup(), passing all of true and iterating over the returned array to filter out the ipv6 addresses if any ipv4 addresses are available:
 
https://nodejs.org/docs/latest-v12.x/api/dns.html#dns_dns_lookup_hostname_options_callback

ehennum:
Here is the Java API equivalent:

https://github.com/marklogic/java-client-api/commit/603cec8124a38e5514f6a3d071158dad5a6c3c5b

georgeajit:
No regressions seen.

PR #571, opened 2021-04-13
#543 - prefer ipv4 to ipv6 addresses

anu3990:
MarkLogic server prefers ipv4 addresses to ipv6 addresses.

Because the client's address factors into validation of the nonce in DIGEST authentication, the client must continue to use the same address while using the nonce.


ehennum (approved):
If the unit tests all run fine with the change, please check in.

Thanks for working through the issues.

anu3990:
All the unit tests ran fine.

Issue #569, opened 2021-04-07
Optic enhancements in 10.0-7


ehennum:
Enhancements:

**1.** rows.generateView() method similar to the existing explain() method:

https://github.com/marklogic/node-client-api/blob/master/lib/rows.js#L294-L344

The signature:

    generateView​(builtPlan, schema, view, queryType)

The schema and view must be non-empty strings.

Only XML is accepted:

    const acceptHeader = 'application/json';

The path sets the output request parameter to "generateView" instead of "explain":

    requestOptions.path = '/v1/rows?output=generateView';

The request appends schemaName and viewName parameters in the same way as other requests that add parameters.

**2.** ModifyPlan existsJoin() and notExistsJoin() methods similar to the existing joinInner() method; bind() similar to the existing select() method

Given that code generation takes care of the joinInner() interface and implementation, code generation should be able to take care of the syntactically identical existsJoin() and notExistsJoin() methods.

This bug identifies the methods for completeness.

**3.**  The methods supporting "Redaction On Rows" (see the internal wiki page)

The methods names: maskDeterministic(), maskRandom(), redactDatetime(), redactEmail(), redactIpv4(), redactNumber(), redactRegex(), redactUsSsn(), and redactUsPhone().

Each method has the same signature. For example:

    maskDeterministic​(column, options)

where the options are an object with name-value pairs specific to the redaction method.

The plan-builder.js module implements a new RdtExpr class with the same shape as the existing SpellExpr class:

https://github.com/marklogic/node-client-api/blob/master/lib/plan-builder-generated.js#L4391

The RdtExpr class provides a method for each redaction method.  The methods are implemented in a way similar to the existing as() method:

https://github.com/marklogic/node-client-api/blob/master/lib/plan-builder-generated.js#L7481-L7497

with the difference that the paramdefs defines an option parameter as follows instead of a expression parameter:

    ['option', [], false, false]]

It should be possible for all of the redaction methods to share an implementation by passing in the function name, column, and options.

Also in the plan-builder.js module, the constructor for the Builder class adds a new field with a RdtExpr instance:

https://github.com/marklogic/node-client-api/blob/master/lib/plan-builder.js#L82-L84

similar to the fields in the base class:

    this.rdt = (fieldOverrides.rdt !== void 0) ? fieldOverrides.rdt : new RdtExpr();

Finally, the plan-builder.js module exports the RdtExpr class:

[...]

PR #572, opened 2021-04-21
#569 -  Optic enhancements in 10.0-7

anu3990:
Adding generateView() in rows.js

### DIFF
--- a/lib/plan-builder-generated.js
+++ b/lib/plan-builder-generated.js
@@ -4433,7 +4433,6 @@ romanize(...args) {
     return new types.XsString('spell', 'romanize', checkedArgs);
     }
 }
-
 class SqlExpr {
   constructor() {
   }
@@ -8267,7 +8266,7 @@ resolveFunction(...args) {
 }
 
 module.exports = {
-PlanBuilder: PlanBuilder,
+  PlanBuilder: PlanBuilder,
 CtsExpr: CtsExpr,
 FnExpr: FnExpr,
 GeoExpr: GeoExpr,
@@ -8280,8 +8279,8 @@ SpellExpr: SpellExpr,
 SqlExpr: SqlExpr,
 XdmpExpr: XdmpExpr,
 XsExpr: XsExpr,
-Plan:       PlanPlan,
-AccessPlan: PlanAccessPlan,
-PlanColumn: PlanColumn,
-PlanExprCol: PlanExprCol
+  Plan:       PlanPlan,
+  AccessPlan: PlanAccessPlan,
+  PlanColumn: PlanColumn,
+    PlanExprCol: PlanExprCol
 };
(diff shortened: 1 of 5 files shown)


lib/rows.js
```diff
@@ -343,4 +343,52 @@ Rows.prototype.explain = function explainRows(builtPlan, format, queryType) {
...
+  requestOptions.headers = {
+    'Content-Type': contentTypeHeader,
+    'Accept': acceptHeader
+  };
+
+  requestOptions.path = '/v1/rows?output=generateView';
+
+  const operation = new Operation(
+      'explain rows', this.client, requestOptions, 'single', 'single');
+  operation.validStatusCodes  = [200, 404];
```
ehennum, line 382: 404 is a valid code when retrieving rows because the query might not match any data.

In this case, it would be an error if the query couldn't be generated -- also in the explain case, which should only allow 200.

test-basic/plan-processors.js
```diff
@@ -314,4 +317,14 @@ describe('processor', function() {
...
+      assert.throws(function () { generateView(
+          p.fromLiterals([
+              {id:1, val: 2},
+              {id:2, val: 4}
+          ]),
+          '', 'xml', ''
+      );}, Error, 'schema cannot be null or empty');
+      done();
+  });
 });
```
ehennum, line 330: It will be good to add a few unit tests for the redaction functions as well, but that will need to wait until I generate the bind() function for the Node.js API.  Just for background, here are the unit tests on the enode:

http://svn.marklogic.com/trac.engsvn/browser/xdmp/trunk/src/jstest/optictest/test-basic/redact.js#L207

Similarly, after I add the enode documentation for the redaction functions, it will be good to add links from the JSDoc for the Node.js client redaction functions.

ehennum (approved):
This revision looks quite close to done.  

The main concern is moving the RdtExpr class to the plan-builder.js module

Thanks for puzzling through the pieces.

lib/plan-builder-generated.js
```diff
@@ -8279,7 +8280,8 @@ SpellExpr: SpellExpr,
 SqlExpr: SqlExpr,
 XdmpExpr: XdmpExpr,
 XsExpr: XsExpr,
-  Plan:       PlanPlan,
-  AccessPlan: PlanAccessPlan,
-  PlanColumn: PlanColumn
+Plan:       PlanPlan,
+AccessPlan: PlanAccessPlan,
+PlanColumn: PlanColumn,
+PlanExprCol: PlanExprCol
```
ehennum, line 8286: Let's just revert all of the changes to plan-builder-generated.js

They're just indentation, so it's better to avoid adding an entry on the version stream for this module.

ehennum (approved):
Looks good after reverting plan-builder-generated.js

Issue #570, opened 2021-04-08
checkConnection() on DatabaseClient is documented as testConnection


wooldridge:
See: 

https://docs.marklogic.com/jsdoc/DatabaseClient.html#testConnection
https://github.com/marklogic/node-client-api/blob/master/lib/marklogic.js#L378

The JSDoc information for the method needs updating.

georgeajit:
Verified.

PR #574, opened 2021-04-28
#570 - Changing testConnection to testConnection

anu3990:
Fix for issue - https://github.com/marklogic/node-client-api/issues/570


ehennum (approved):
Thanks for fixing that.

Issue #485, opened 2017-12-13
Throw error when complexValues is set with a chunked stream


wooldridge:
Make it clear to developers that this is not allowed. See #484.

georgeajit:
Unit test is sufficient and runs fine.

PR #575, opened 2021-04-28
#485 - Throw error when complexValues is set with a chunked stream

anu3990:
Fix for issue - https://github.com/marklogic/node-client-api/issues/485


ehennum (approved):
Thanks for the fast turnaround!

PR #577, opened 2021-05-10
Test commit - dynamic interface to Data Service endpoints

anu3990:


anu3990:
Pinging @georgeajit for awareness.

lib/endpoint-proxy.js
```diff
@@ -521,11 +768,70 @@ class EndpointProxy {
...
+      throw new Error('no database client provided');
+    }
+    if (serviceDeclaration === void 0 || serviceDeclaration === null) {
+      throw new Error('no service declaration provided');
+    }
+    const db = client.getConnectionParams().database;
+    if (db !== void 0 && db !== null) {
+      throw new Error('connection for client cannot specify database');
+    }
+    this._client = client;
```
ehennum, line 783: Because the $mlProxy field keeps a reference to the client, I believe it's not necessary for this class to keep a client field.

lib/endpoint-proxy.js
```diff
@@ -521,11 +768,70 @@ class EndpointProxy {
...
+    const overrideExtension = this._moduleExtension;
+    const funcdef = new Functiondef(
+        functionDeclaration,
+        this._endpointDirectory,
+        (overrideExtension !== null) ? overrideExtension : moduleExtension
+    );
+
+    this._functiondefs.set(funcdef.name(), funcdef);
+    return this;
+  }
```
ehennum, line 796: The issue was trying to suggest that the constructor should call withFunction() and not that the class needs to provide a withFunction() implementation.

ehennum (commented):
Thanks for digging into it.  

I provided some suggestions for the constructors. Please see what you think.

Issue #567, opened 2021-04-06
dynamic interface to Data Service endpoints


ehennum:
Add a new endpointCaller() factory method to the DatabaseClient object with the following signature:

    endpointCaller(serviceDeclaration, endpointDeclarations) -> EndpointProxy

The parameters:

    serviceDeclaration = an optional object with endpointDirectory and endpointExtension properties
         similar to the com.marklogic.client.impl.BaseProxy constructor in Java

    endpointDeclarations = a required array with objects equivalent to the *.api JSON structure having
        either an endpoint property with the full path or a functionName property for use with endpointDirectory and endpointExtension
        an optional params property
        an optional return property

The endpointCaller() factory method of the client passes the arguments to a new initCaller() factory function exported by the lib/endpoint-proxy.js module.

The initCaller() factory function does the equivalent of 

- constructing a new instance of the existing EndpointProxy class and
- calling the existing withFunction() function for each endpoint declaration
- returning the new instance of the EndpointProxy class

The existing EndpointProxy class changes as needed to support both the existing statically generated Data Service endpoint class (with full backward compatibility) and the new dynamically constructed endpoint caller.

In particular, the serviceDeclaration provides a default endpointDirectory and endpointExtension if the endpointDeclaration specifies a functionName instead of an endpoint. 

Also, lines 30 through 275 of the lib/proxy-generator.js module probably need to move into the lib/endpoint-proxy.js module so the expandEndpointDeclaration() function is available for preparing a dynamically constructed endpoint caller.  

If so, lib/endpoint-proxy.js would need to export the expandEndpointDeclaration() function so the lib/proxy-generator.js module can require lib/endpoint-proxy.js and still call the expandEndpointDeclaration() function after the refactoring.

Finally, the constructed instance will need to provide an interface for calling the execute() method of the EndpointProxy class on each endpoint with arguments, returning a selector for whether the response is provided as a promise or stream (in the usual way for the Node.js API).

PR #578, opened 2021-05-12
#567 - dynamic interface to Data Service endpoints

anu3990:
Dynamic interface to Data Service endpoints


lib/endpoint-proxy.js
```diff
@@ -521,11 +768,92 @@ class EndpointProxy {
...
+    if (endpointDeclaration === void 0 || endpointDeclaration === null) {
+      throw new Error('endpointDeclaration cannot be null.');
+    }
+    const endpoint = endpointDeclaration.endpoint;
+
+    if (endpoint === void 0 || endpoint === null) {
+      throw new Error('endpointDeclaration endpoint cannot be null.');
+    }
+    const regex = /^(.*[/])([^/]+)(.[^/.]+)$/;
+    const arrayValues = endpoint.match(regex);
```
ehennum, line 818: Would it be defensive to test that the array has at least 4 items (the [0] item is the entire match; the other items are the matching groups)?  

That way, input that doesn't match the regex wouldn't trigger an array access error on line 820?

Also, would endpointFields be a more semantic name and thus easier to understand than arrayValues (which is syntactic)?  

By the way, the regex was just a guess -- if you haven't done so, it would be a good idea to execute the match() expression in Query Console or any other JavaScript environment on some paths that should be expected to succeed or fail, as in:

- /dir/file.extension - should succeed
-  /file.extension - should succeed
- file.extension - should fail
- /dir/file - should fail
- /dir/.extension - should fail

lib/endpoint-proxy.js
```diff
@@ -521,11 +768,92 @@ class EndpointProxy {
...
+
+    const endpointDirectory = arrayValues[1];
+    const endpointFunctionName = arrayValues[2];
+    const endpointExtension = arrayValues[3];
+
+    if(endpointDirectory == null || endpointExtension == null) {
```
ehennum, line 824: Would it be defensive to check that each of endpointDirectory, endpointFunctionName, and endpointExtension is a string primitive or object (similar to the test of the endpoint property above)?

lib/endpoint-proxy.js
```diff
@@ -521,11 +768,92 @@ class EndpointProxy {
...
+      throw new Error('endpoint must contain endpointDirectory and extension');
+    }
+    this.functionName = endpointFunctionName;
+    endpointDeclaration.functionName = this.functionName;
+
+    const serviceDeclaration = mlutil.copyProperties(endpointDeclaration);
+    serviceDeclaration.functionName = endpointFunctionName;
```
ehennum, line 831: Sorry, I think my previous suggestion was unclear.

For the service declaration, I believe it should work to construct a literal object:

`    {endpointDirectory: endpointDirectory, endpointExtension: endpointExtension}`

For the new endpoint declaration, the suggestion was to copy the input endpoint declaration and set the functionName on the copy (not on the input) so the code doesn't have the side effect of modifying the input.  (In other words, line 828 should be avoided.)

It's a good practice to avoid modifying input directly because the application may be using the input for other purposes.

lib/marklogic.js
```diff
@@ -716,6 +716,20 @@ MarkLogicClient.prototype.createProxy = function createProxy(serviceDeclaration)
   return proxy.init(this, serviceDeclaration);
 };
 
+MarkLogicClient.prototype.serviceCaller = function serviceCaller(serviceDeclaration, endpointDeclarations) {
+  if(serviceDeclaration == null || endpointDeclarations == null) {
+    throw new Error('Parameters - serviceDeclaration and endpointDeclarations are required');
+  }
```
ehennum, line 722: Given that the ServiceCaller constructor validates, it's might be unnecessary to validate here.

lib/marklogic.js
```diff
@@ -716,6 +716,20 @@ MarkLogicClient.prototype.createProxy = function createProxy(serviceDeclaration)
...
+  return proxy.initService(this, serviceDeclaration, endpointDeclarations);
+};
+
+MarkLogicClient.prototype.endpointCaller =function endpointCaller(endpointDeclaration) {
+  if(endpointDeclaration == null) {
+    throw new Error('endpointDeclaration cannot be null');
+  }
```
ehennum, line 729: Given that the EndpointCaller constructor validates, it's might be unnecessary to validate here.

lib/marklogic.js
```diff
@@ -844,6 +858,8 @@ function prefLookup(hostname, options, callback) {
   dns.lookup(hostname, {all:true, verbatim:false}, prefCallback);
 }
 
+
+
```
ehennum, line 862: Best to remove this change so the diff has only significant changes?

ehennum (approved):
A few additional suggestions -- other than a clarification about the construct service and endpoint declarations in EndpointCaller, the suggestions are mostly about defensive programming.

After these fixes, I think the only remaining step is the unit tests.

Issue #579, opened 2021-05-12
Jenkins failure - Tests in probe.js


anu3990:
Two tests in probe.js with title "probe in transaction using param list.should find written doc in same transaction"-  throw 403 exception in Jenkins. The user - rest-writer need xdmp-invoke privilege in order to run these tests. 
As a fix, rest-evaluator role (this internally has xdmp-invoke priviledge), needs to be added to the user - rest-writer

georgeajit:
No failures seen.

PR #580, opened 2021-05-12
#579 - Jenkins failure - Tests in probe.js

anu3990:
Adding rest-evaluator role to rest-writer.

### DIFF
--- a/etc/test-setup-users.js
+++ b/etc/test-setup-users.js
@@ -81,6 +81,18 @@ function setupUsers(manager, done) {
       }).result();
   }).
   then(function(response) {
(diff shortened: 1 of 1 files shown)


ehennum (changes requested):
I suspect that a GET request merely reads without making changes.

I suspect it will be necessary to add the rest-evaluator role to the rest-writer and maybe even rest-reader users.

I suspect that the diff showing work Ling did on the setup for the Optic API tests may provide an example.

ehennum (approved):
Looks reasonable to me -- thanks for taking care of that issue.

PR #581, opened 2021-05-20
Fix for jenkins failure.

anu3990:


ehennum (approved):
Thanks for the fine tuning

PR #582, opened 2021-05-24
Unit tests for dynamic interface to Data Service endpoints

anu3990:


ehennum (approved):
Looks great -- thanks for putting a clean interface for dynamic use of the refactored implementation.

My only suggestion is to add a test for the endpointCaller()

I believe that could be done by copying one of the existing tests and making the following changes in the copy:

*  instead of setting endpointExtension on the service declaration, sets endpoint on the endpointDeclaration to serviceDeclaration.endpointDirectory + endpointDeclaration.functionName + THE_EXTENSION_FOR_THIS_ENDPOINT
* instead of calling dbWriter.serviceCaller() with the serviceDeclaration and endpointDeclaration, calls dbWriter.endpointCaller() with the endpointDeclaration
* executes the call() without the function name, passing only the params

Everything else in the copied test should be fine.

One other thought -- to verify support for multiple endpoint declarations, it would be good to pass an array with two endpoint declarations to a serviceCaller() and chain the calls, as in:

    serviceCaller.call(...first function and params...)
       .then(ouput => {
            ... assertions for first function but no call to done() ...
            return serviceCaller.call(...second function and params...);
        })
       .then(ouput => {
            ... assertions for second function ...
             done();
        })
       .catch(...as usual...);

Returning the second call returns a Promise, so the chain can continue.

anu3990:
> 
> I believe that could be done by copying one of the existing tests and making the following changes in the copy:
> 
> * instead of setting endpointExtension on the service declaration, sets endpoint on the endpointDeclaration to serviceDeclaration.endpointDirectory + endpointDeclaration.functionName + THE_EXTENSION_FOR_THIS_ENDPOINT
> * instead of calling dbWriter.serviceCaller() with the serviceDeclaration and endpointDeclaration, calls dbWriter.endpointCaller() with the endpointDeclaration
> * executes the call() without the function name, passing only the params
> 
> Everything else in the copied test should be fine.
> 
> One other thought -- to verify support for multiple endpoint declarations, it would be good to pass an array with two endpoint declarations to a serviceCaller() and chain the calls, as in:
> 
> ```
> serviceCaller.call(...first function and params...)
>    .then(ouput => {
>         ... assertions for first function but no call to done() ...
>         return serviceCaller.call(...second function and params...);
>     })
>    .then(ouput => {
>         ... assertions for second function ...
>          done();
>     })
>    .catch(...as usual...);
> ```
> 
> Returning the second call returns a Promise, so the chain can continue.

Created a new pull request with additional unit tests.

anu3990:
Thanks @ehennum.! Created a new pull request with the above mentioned suggestions - https://github.com/marklogic/node-client-api/pull/583

PR #583, opened 2021-05-25
Endpoint Caller unit tests.

anu3990:
Adding unit tests for Endpoint caller and an additional test for serviceCaller

### DIFF
--- a/test-basic/endpoint-caller.js
+++ b/test-basic/endpoint-caller.js
@@ -20,16 +20,16 @@ var testconfig = require('../etc/test-config.js');
 var expect = require('chai').expect;
 var dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnection);
 
-describe('Service caller', function() {
+describe('Endpoint caller', function() {
 
     it('moduleInitSjs endpoint', function(done) {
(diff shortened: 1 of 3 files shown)


test-basic/endpoint-caller.js
```diff
@@ -0,0 +1,100 @@
...
+var testconfig = require('../etc/test-config.js');
+var expect = require('chai').expect;
+var dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnection);
+
+describe('Endpoint caller', function() {
+
+    it('moduleInitSjs endpoint', function(done) {
+        const serviceDeclaration = JSON.parse(fs.readFileSync('test-basic-proxy/ml-modules/generated/moduleInitSjs/service.json', {encoding: 'utf8'}));
+        const endpointDeclaration = JSON.parse(fs.readFileSync('test-basic-proxy/ml-modules/generated/moduleInitSjs/initializer.api', {encoding: 'utf8'}));
+        endpointDeclaration.endpoint = serviceDeclaration.endpointDirectory+'initializer'+'.sjs';
```
ehennum, line 28: A potential small refinement would be to use endpointDeclaration.functionName instead of the literal quoted string when setting the endpointDeclaration.endpoint property here and in the other tests.

That way, it's just a little easier to understand how the pieces fit together and a little easier to maintain.

ehennum (approved):
Looks great to me.  Thanks for establishing a reliable baseline -- and examples -- for the capability.

PR #584, opened 2021-05-26
Update NOTICE.txt

anu3990:
https://project.marklogic.com/jira/browse/RC-2425


ehennum (approved):
In updating the NOTICE.txt, we rely on legal for the content.

That is, if we've replaced all of the existing text with the new text, and the text is readable, our part is done.

PR #585, opened 2021-06-02
Unit tests for redaction and generate-view.

anu3990:
Unit tests for redaction and generate-view

### DIFF
--- a/lib/plan-builder.js
+++ b/lib/plan-builder.js
@@ -118,7 +118,9 @@ class RdtExpr {
    * Controls features such as the length and type of the generated value.
    * @method planBuilder#maskDeterministic
    * @since 2.7.0
-   * @param args - Options for redacting.
+   * @param args - with column and options - Column is the name of the column to be defined.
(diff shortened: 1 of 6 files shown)


lib/plan-builder.js
```diff
@@ -115,114 +118,96 @@ class RdtExpr {
    * Controls features such as the length and type of the generated value.
    * @method planBuilder#maskDeterministic
    * @since 2.7.0
-   * @param { PlanColumnName } [column] - The name of the column to be defined.
-   *  This can be either a string or the return value from PlanBuilder.col(), PlanBuilder.viewCol() or PlanBuilder.schemaCol().
-   * @param { Item } [options] - An object with name-value pairs specific to the redaction method.
```
ehennum, line 120: In the JSDoc, the previous documentation of the params is correct both here and in the other public redaction functions:  the caller passes the column and option params.

The ...args syntax is a JavaScript shortcut for setting an array with the first item from the column param and the second item from the options param.

ehennum (approved):
My only concern is with the JSDoc for the parameters.

(The JavaScript syntax hides the fact that the caller still passes two parameters, as in the unit tests.)

After reverting to the old parameter JSDoc, please merge.

PR #589, opened 2021-07-27
NODE-26 : https://project.marklogic.com/jira/browse/NODE-26 (Draft)

anu3990:
Initial prototype for ingestion - https://project.marklogic.com/jira/browse/NODE-26 using Data Movement Api.

### DIFF
--- a/test-basic/concurrent-requests.js
+++ b/test-basic/concurrent-requests.js
@@ -39,7 +39,7 @@ function writeBatch(writerId, nextDocument) {
         return;
     }
     var documentDescriptors = createDocumentDescriptors();
-    console.log('wrierId is '+writerId);
+    console.log('writerId is '+writerId);
     console.log('writing batch '+nextDocument+' to '+(nextDocument+10));
     var batch = documentDescriptors.slice(nextDocument, nextDocument +10);
     nextDocument = nextDocument+10;
--- a/test-basic/concurrent-requests.js
+++ b/test-basic/concurrent-requests.js
@@ -14,48 +14,48 @@
  * limitations under the License.
  */
 
-var testconfig = require('../etc/test-config.js');
+const testconfig = require('../etc/test-config.js');
 
-var marklogic = require('../');
+const marklogic = require('../');
 
-var dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnection);
+const dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnection);
 
 describe('concurrent-requests test', function(){
 
 
     it('should write 1000 documents concurrently', function(done){
         for(var i=0; i<3; i++) {
-            writeBatch(i,0);
+            writeBatch(i,done);
         }
 
-        done();
     });
 });
 
-
-function writeBatch(writerId, nextDocument) {
+const documentDescriptors = createDocumentDescriptors();
+let nextDocument = 0;
+function writeBatch(writerId, done) {
     if(nextDocument>=1000) {
         console.log(writerId + ' is done');
-        return;
+        return done();
     }
-    var documentDescriptors = createDocumentDescriptors();
     console.log('writerId is '+writerId);
     console.log('writing batch '+nextDocument+' to '+(nextDocument+10));
-    var batch = documentDescriptors.slice(nextDocument, nextDocument +10);
+    const batch = documentDescriptors.slice(nextDocument, nextDocument +10);
     nextDocument = nextDocument+10;
 
-    dbWriter.documents.write(batch).result(writeBatch(writerId, nextDocument)).catch();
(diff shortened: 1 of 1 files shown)


test-basic/concurrent-requests.js
```diff
@@ -0,0 +1,64 @@
...
+
+describe('concurrent-requests test', function(){
+
+
+    it('should write 1000 documents concurrently', function(done){
+        for(var i=0; i<3; i++) {
+            writeBatch(i,0);
+        }
+
+        done();
```
ehennum, line 31: Part of the issue might be on line 31.  Currently done() gets called after the loop finishes starting the 3 writers, but done() should actually be called after after the last batch is written.

Unlike most Java, the call stack in Node.js doesn't reflect the sequence of execution.  That is, the response callbacks execute after the code that passes in the callbacks finishes.

I'd suggest passing the done function reference as another parameter to writeBatch().

test-basic/concurrent-requests.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+});
+
+
+function writeBatch(writerId, nextDocument) {
+    if(nextDocument>=1000) {
+        console.log(writerId + ' is done');
```
ehennum, line 38: I'd suggest calling done() here when all of the writes have been queued.  That's sufficient for the prototype.

In the real version, the code will need to notify the caller only when the writes have finished.

test-basic/concurrent-requests.js
```diff
@@ -0,0 +1,64 @@
...
+        return;
+    }
+    var documentDescriptors = createDocumentDescriptors();
+    console.log('writerId is '+writerId);
+    console.log('writing batch '+nextDocument+' to '+(nextDocument+10));
+    var batch = documentDescriptors.slice(nextDocument, nextDocument +10);
+    nextDocument = nextDocument+10;
+
+    dbWriter.documents.write(batch).result(writeBatch(writerId, nextDocument)).catch();
```
ehennum, line 47: Another part of the issue might be here.  Line 47 calls writeBatch() immediately and passes the return value to result.

In other words, the calls to writeBatch are recursing here such that the 1st iteration of the for loop processes all of the items in the array.  There's nothing left for the other 2 iterations of the loop.

Instead, line 47 should pass a callback that executes when the write() request finishes.  That's what allows other requests to execute.

Creating and passing a lambda function should work:

    .result((output) => writeBatch(writerId, nextDocument, done))

Also, .catch() should at a minimum provide a logging callback such as

    .catch((err) => console.log(err));

Otherwise, the error is just silently ignored -- similar to a Java catch clause that does nothing with the caught exception.

test-basic/concurrent-requests.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function createDocumentDescriptors() {
+    var documentDescriptors = [];
+
+    for(var i=0; i<1000; i++) {
+        var temp = {
+            uri: '/test/concurrent/requests'+i+'.json',
+            contentType: 'application/json',
+            content: new Buffer('{"key"'+i+':"value'+i+' 1"}')
```
ehennum, line 58: In Node.js 10, I believe it should also work to supply the content as an in-memory JavaScript literal object:

    content: {[ 'key'+i]: 'value'+i+' 1'}

ehennum (changes requested):
I suspect the main issue that the test calls the response action immediately instead of passing a reference to a callback function for a deferred call.

test-basic/concurrent-requests.js
```diff
@@ -0,0 +1,64 @@
...
+    });
+});
+
+
+function writeBatch(writerId, nextDocument) {
+    if(nextDocument>=1000) {
+        console.log(writerId + ' is done');
+        return;
+    }
+    var documentDescriptors = createDocumentDescriptors();
```
ehennum, line 41: The documentDescriptors array and the nextDocument counter should be shared across all calls to writeBatch().

I'd suggest turning the documentDescriptors variable into a module-level constant by moving the declaration immediately before the start of the writeBatch() function on line 36.

Also, I was mistaken to suggest that nextDocument should be a parameter.  I'd suggest turning it into a module-level variable.

Putting the two suggestions together, the declaration would look like:

    const documentDescriptors = createDocumentDescriptors();
    let nextDocument = 0;
    function writeBatch(writerId) {
        ...
    }

By the way, the var keyword is discouraged in recent versions of JavaScript because it is not properly scoped in execution.  The recommendation is to use const to declare immutable constants (equivalent to final in Java) and let to define mutable variables.

ehennum (approved):
Phase 1 of the prototype done -- on to Phase 2 (reading from a stream)!

The prototype will likely evolve to become the data movement function for document ingestion, at which point it would make sense to merge into the Node.js API GitHub repository.

PR #590, opened 2021-08-12
NODE-27 : Prototype for ingestion using NodeJs streams

anu3990:


test-basic/data-movement-request.js
```diff
@@ -0,0 +1,64 @@
...
+function writeDocs(record, done) {
+    if(record === null) {
+        return done();
+    }
+
+    dbWriter.documents.write(record).result((output) => writeDocs(readable.read(), done))
+        .catch((err) => console.log(err));
+
+}
+
```
ehennum, line 64: Does the current test prototype the approach from the wiki where the application writes to a stream and the API implementation reads from the same stream?

Specifically, I would expect the prototype to do the following:

- write documents to a stream.Writable (possibly a stream.PassThrough)
- read documents from a stream.Readable (possibly the same stream.PassThrough)
- buffer a batch of documents in an array
- submit the batch when the array is full or the stream ends 
- have multiple concurrent writes pending at the same time (similar to the previous prototype)
- call done when the response from the last write returns

ehennum (changes requested):

test-basic/data-movement-request.js
anu3990: Hi Erik
The stream "readable" is the stream where data is written to and read from. "createDocumentDescriptors()" writes to this stream (pushes the document descriptors) and then "writeDocs()" reads from this stream and writes to database. So the changes will be - changing the type of "readable" to stream.PassThrough and the batch implementation?

ehennum: Sorry for the lack of clarity.  The purpose of prototyping is to make progress toward the implementation.

The first prototype supported reading batches of documents from an array by independent requests that processed each response by making a new request.

To make progress toward the implementation, it would be great if the second prototype built on the first.  The change would be to read from a stream instead of an array.  Management of concurrency should be similar to the first prototype -- that's the critical aspect of a data movement function.  

How the stream is populated isn't that important.  Ideally, the stream would be populated by writing to a stream.PassThrough, but it's also possible to wrap a stream.Readable implementation around an array.

test-basic/data-movement-request.js
```diff
@@ -50,15 +45,25 @@ function createDocumentDescriptors() {
...
-        return done();
+function writeDocs(writerId, done) {
+
+    const writeBatchArray = [];
+    for(let i=0; i<batchSize; i++) {
+        const record = readable.read();
+        if(record === null) {
+            if(writeBatchArray.length!==0) {
+                dbWriter.documents.write(writeBatchArray);
+            }
```
ehennum, line 59: If the last batch is partial, will writeDocs() return before it is written?

Instead, would it work to break out of the loop instead of returning and have a branch after the loop similar to the following:

    switch(writeBatchArray.length) {
    case 0: ...call done() immediately and break...
    case batchSize:  ...call documents.write() with the writeDocs() callback and break...
    default: ...call documents.write() with done() as the callback and break...
    }

test-basic/data-movement-request.js
```diff
@@ -20,26 +20,21 @@ const dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnection)
...
-        readable.pipe(writable);
-        createDocumentDescriptors();
-        readable.on('readable', function() {
-            writeDocs(readable.read(), done);
-        });
+        createInput();
+        for(var i=0; i<3; i++) {
+            writeDocs(i, done);
+        }
         done();
```
ehennum, line 33: Calling done() after starting the requests means the test will finish before the stream is read.

Instead, done() should be called only after the last batch finishes writing.

ehennum (approved):
One suggestion for handling a partial last batch and calling done() after writing finishes instead of after writing is initialized.

Other than that, it looks ready to evolve into the new writeAll() function in the documents.js module.

test-basic/data-movement-request.js
anu3990: Done

test-basic/data-movement-request.js
anu3990: Hi Erik, I did not remove this done() since moving this makes the test throw the error - 
Error: done() called multiple times in test <data-movement-requests test should write 1000 documents> of file

ehennum: > Error: done() called multiple times in test

That indicates that the finalization after the last response isn't working correctly because it is calling done() more than once. 

Calling done() after the loop on line 38 finishes the test before the first request has been sent to the server.

In a prototype, such problems don't matter much but

*  In the real implementation, it will be important to finalize only after the last response.
*  In the real test, it will be necessary to call done() only after the last response finishes so that the final result can be inspected.
 
Whether to work out the fix now in the prototype or in the real implementation is up to the implementer...

PR #600, opened 2021-09-02
Node 9: Operation

anu3990:


lib/documents.js
```diff
@@ -2266,9 +2267,39 @@ function writeAllDocumentsImpl(options) {
...
+      }
+    }
+    writeBatchArray.push(record);
+  }
+
+  if(writeBatchArray.length !== 0) {
+    writeDocumentList.call(this, writeBatchArray, requestOptions).result((output) => writeAllDocumentsImpl(writerId, done))
+        .catch((err) => console.log(err));
+  }
+
```
ehennum, line 2292: This new code must execute when it is time to write a new batch of documents and not when the initialization request starts. 

I'd suggest implementing a new onWriteAllInit() function.

Pass a reference to the new function as the callback for the initialization response, as in:

`requester.startRequest(operation).result(mlutil.callbackOn(operation, onWriteAllInit));
`

Eventually, it will be important to pass an error handler as the second argument to result(), but that can wait.

The mlutil.callbackOn() ensures that this is a reference to the operation object with the state.

Suggestions for the body of the onWriteAllInit(output) function:

```
function onWriteAllInit(output) {
    const operation = this;
   // get the number of hosts and forests from the output
   // set the number of hosts and forests in the operation
   // where the phase 2 prototype used 3 for the maximum number of requesters,
   //     instead use the number of hosts for now; and a configured multiple eventually
   for (let i=0; i < maxRequesters; i++ {
       // start the requests as in the phase 2 prototype, but pass the operation object
   }
}
```

The code added in lines 2270 through 2290 above belongs in function that's called from the for loop in onWriteAllInit() and from the callback for the writeDocumentList() result.

Use the operation object instead of this.

lib/documents.js
```diff
@@ -2248,7 +2249,7 @@ Documents.prototype.writeAll = function writeAllDocuments(options) {
       this, options
   );
 };
-function writeAllDocumentsImpl(options) {
+function writeAllDocumentsImpl(options, done) {
```
ehennum, line 2252: done is a test function, not a production application function

The test will need to configure the options with job success and failure callbacks.

In those callbacks, the test will call done().

lib/documents.js
```diff
@@ -2266,9 +2267,39 @@ function writeAllDocumentsImpl(options) {
...
+
+  if(writeBatchArray.length !== 0) {
+    writeDocumentList.call(this, writeBatchArray, requestOptions).result((output) => writeAllDocumentsImpl(writerId, done))
+        .catch((err) => console.log(err));
+  }
+
   return inputStream;
 }
 
+function finishWriter(writerId, done) {
```
ehennum, line 2296: This function should take the operation object instead of done and should access the requesterCount as a property of the operation object.

ehennum (changes requested):
When working with Node.js or any other asynchronous environment, it's important to realize that the response processing goes in a callback instead of following the request.

lib/documents.js
```diff
@@ -2263,12 +2264,143 @@ function writeAllDocumentsImpl(options) {
...
+
+  if(operation.jobState.requesterCount<1) {
+
+    if(operation.jobState.jobOptions && operation.jobState.jobOptions.onCompletion){
+
+      let summary = ('Number of documents written successfully = '+
+          ((operation.jobState.docsWrittenSuccessfully>0)?operation.jobState.docsWrittenSuccessfully:0)+'\n'+
+          'Number of documents failed to be written = '+
+          ((operation.jobState.docsFailedToBeWritten>0)?operation.jobState.docsFailedToBeWritten:0)
+      );
```
ehennum, line 2296: Instead of assembling a string, it would be more useful to provide the application with an object that has a property for each value, similar to the following sketch:

```
    const summary = {
        docsWritten: ...,
        docsFailed: ...
        };
     if (jobState.error) {
        summary.error = ... ;
    }
```

lib/documents.js
```diff
@@ -2263,12 +2264,143 @@ function writeAllDocumentsImpl(options) {
...
+      jobState.jobOptions.batchSize >0) ?
+      jobState.jobOptions.batchSize:100;
+
+  a:
+      for (let i = 0; i < batchSize; i++) {
+        const record = jobState.stream.read();
+
+        if(jobState.jobOptions && jobState.jobOptions.defaultMetadata) {
+          writeBatchArray.push(jobState.jobOptions.defaultMetadata);
+        }
```
ehennum, line 2349: Yes, that's the intent of jobOptions.defaultMetadata

It will be necessary to validate the properties of the defaultMetadata descriptor in writeAllDocumentsImpl().  The metadata descriptor can take the same properties as the document descriptor except uri and content.

lib/documents.js
```diff
@@ -2263,12 +2264,143 @@ function writeAllDocumentsImpl(options) {
...
+          })
+        .catch((err) => {
+          jobState.docsFailedToBeWritten+= writeBatchArray.length;
+          if(jobState.jobOptions && jobState.jobOptions.onBatchError) {
+            const progressSoFar = {
+              docsWrittenSuccessfully: jobState.docsWrittenSuccessfully,
+              docsFailedToBeWritten: jobState.docsFailedToBeWritten
+            };
+            const documents = writeBatchArray;
+            jobState.jobOptions.onBatchError(progressSoFar, documents, err);
```
ehennum, line 2396: This is a good start for the error notification.  

As in Java, it will be necessary to inspect the return value to determine whether to retry the batch, skip the batch, or end the stream.

lib/documents.js
```diff
@@ -2263,12 +2264,143 @@ function writeAllDocumentsImpl(options) {
...
+              writeDocumentsImpl.call(jobState.docInstance, false, writeBatchArray)
+                  .result((out) =>{
+                    jobState.docsWrittenSuccessfully+= writeBatchArray.length;
+                      finishWriter(operation);
+                      })
+                  .catch((err) => {
+                    jobState.docsFailedToBeWritten+= writeBatchArray.length;
+                    jobState.error = err;
+                    jobState.stream.emit('error', err);
+                  });
```
ehennum, line 2366: Given that the call to and result processing for writeDocumentsImpl() is the same on line 2357 and 2375 except for the last call in the success and error handlers, maybe it's time to eliminate the duplicate code.  

One possibility would be for the default case to set a flag that can be tested to guard the call to finishWriter() after the call to onBatchSuccess() -- in other words, after current line 2385

ehennum (commented):
This revision is going in a good direction.

Unit tests for the possible job options will be helpful.

ehennum (commented):
I would suggest providing only the data instead of any text.  

That way, that application doesn't have to strip out any text to get the data.

For example:

    elapsedMilliseconds:  jobState.timeElapsed

    summary.error = jobState.error;

lib/documents.js
ehennum: If I recall correctly, a later commit switched from the string to an object.

Unfortunately, I haven'f found a way to get GitHub to bundle up the commits as a single diff so the review can be more consistent.  Have you?

lib/documents.js
```diff
@@ -2265,105 +2265,123 @@ function writeAllDocumentsImpl(jobOptions) {
...
         const record = jobState.stream.read();
 
-        if(jobState.jobOptions && jobState.jobOptions.defaultMetadata) {
-          writeBatchArray.push(jobState.jobOptions.defaultMetadata);
+        if(record !== null) {
+          if (jobState.jobOptions.defaultMetadata) {
+            writeBatchArray.push(jobState.jobOptions.defaultMetadata);
+          }
+          if (jobState.jobOptions.transform) {
+            writeBatchArray.push(jobState.jobOptions.transform);
```
ehennum, line 2375: The transform should passed on the call to writeDocumentsImpl() -- see the JSDoc for writeDocuments()

It might be more convenient to use the object instead of the positional form of the call as in:

```
const batchdef = {documents: writeBatchArray};
if (jobState.jobOptions.transform) {
    batchdef.transform = jobState.jobOptions.transform;
}
writeDocumentsImpl.call(jobState.docInstance, false, batchdef)

```

ehennum (commented):
This looks like definite progress to me and a basis for unit testing that confirms the robustness of the work so far.  Then, the other properties from the story can be added.

lib/documents.js
```diff
@@ -2265,105 +2265,123 @@ function writeAllDocumentsImpl(jobOptions) {
...
+ }
+
   let jobState = {
     docInstance : this,
     stream: inputStream,
     requesterCount: 0,
     docsWrittenSuccessfully:0,
     docsFailedToBeWritten: 0,
-    jobOptions: jobOptions
+    jobOptions: (jobOptions)? mlutil.copyProperties(jobOptions):{}
```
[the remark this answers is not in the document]
anu3990: Done. Moved jobState initialization before setting the defaults.

lib/documents.js
anu3990: Done.

lib/documents.js
```diff
@@ -2248,12 +2248,17 @@ Documents.prototype.writeAll = function writeAllDocuments(options) {
       this, options
   );
 };
-function writeAllDocumentsImpl(options) {
+function writeAllDocumentsImpl(jobOptions) {
+
+  if(!jobOptions) {
+    throw new Error('jobOptions cannot be null or undefined.');
+  }
```
ehennum, line 2255: The story claims that jobOptions can be optional.  Line 2278 tolerates job options that are optional.  

Is the check on line 2253 - 2255 appropriate?

lib/documents.js
```diff
@@ -2263,12 +2268,184 @@ function writeAllDocumentsImpl(options) {
...
+  let jobState = {
+    docInstance : this,
+    stream: inputStream,
+    requesterCount: 0,
+    docsWrittenSuccessfully:0,
+    docsFailedToBeWritten: 0,
+    jobOptions: (jobOptions)? mlutil.copyProperties(jobOptions):{}
+  };
+
+ if(jobState.jobOptions){
```
ehennum, line 2281: Line 2278 ensures jobState.jobOptions will always be an object.

Can the test on jobState.jobOptions that guards this branch be removed such that the branch becomes part of the main flow.

lib/documents.js
```diff
@@ -2263,12 +2268,184 @@ function writeAllDocumentsImpl(options) {
...
+   }
+   else {
+     jobState.jobOptions.concurrentRequests = {multipleOf:'forests', multiplier:4};
+   }
+
+   // Check for jobOptions.batchSize
+   jobState.jobOptions.batchSize = (jobState.jobOptions.batchSize && jobState.jobOptions.batchSize >0) ? jobState.jobOptions.batchSize:100;
+
+   //Check for defaultMetadata
+   jobState.jobOptions.defaultMetadata = (jobState.jobOptions.defaultMetadata)? jobState.jobOptions.defaultMetadata:{};
```
ehennum, line 2308: Defaulting to an empty metadata object results in sending an empty metadata payload to the server, which adds to the expense of the request without benefit to the application. 

Would it be better to default to null and skip sending the metadata payload if it is null?

lib/documents.js
```diff
@@ -2263,12 +2268,184 @@ function writeAllDocumentsImpl(options) {
...
+  switch (jobState.jobOptions.concurrentRequests.multipleOf) {
+    case 'forests':
+      maxRequesters = jobState.jobOptions.concurrentRequests.multiplier * output.length;
+      break;
+    case 'hosts':
+      maxRequesters = jobState.jobOptions.concurrentRequests.multiplier *
+          new Set(output.map((forest) => forest.host)).size;
+      break;
+    default:
+      throw new Error('Invalid value for onCompletion.multipleOf. Value must be forests or hosts.');
```
ehennum, line 2351: Because this is a deferred callback, an error thrown here won't be received by the application.

Instead, it would be better to emit an error on the stream.

lib/documents.js
```diff
@@ -2263,12 +2268,184 @@ function writeAllDocumentsImpl(options) {
...
+  const batchSize = jobState.jobOptions.batchSize;
+  jobState.lastBatch = false;
+  jobState.retryCount = 2;
+
+  a:
+      for (let i = 0; i < batchSize; i++) {
+        const record = jobState.stream.read();
+
+        if(record !== null) {
+          if (jobState.jobOptions.defaultMetadata.length>0) {
```
ehennum, line 2372: Should this test whether defaultMetadata is set?

Because defaultMetadata is an object instead of an array, it shouldn't have a length property. 

Also, the test should check i === 0 so defaultMetadata is pushed only before the first document.  At present, the default metadata object is being added before each document, which adds to the cost with the same functionality.

lib/documents.js
```diff
@@ -2263,12 +2268,184 @@ function writeAllDocumentsImpl(options) {
...
+          })
+          .catch((err) => {
+              if (jobState.jobOptions.onBatchError) {
+                const progressSoFar = {
+                  docsWrittenSuccessfully: jobState.docsWrittenSuccessfully,
+                  docsFailedToBeWritten: writeBatchArray.length,
+                  timeElapsed: (Date.now()-jobState.startTime)
+                };
+                const documents = writeBatchArray;
+                jobState.jobOptions.onBatchError(progressSoFar, documents, err);
```
ehennum, line 2424: A future iteration will need to process the return value from onBatchError() as described in the story, but that iteration can occur after this iteration is merged.

ehennum (commented):
Good progress and a basis for unit tests.

lib/documents.js
anu3990: Thanks Erik.! Changed the same for "transform" as well.

lib/documents.js
anu3990: Changed the same for "transform" as well. Should transform also be added only if i===0?

lib/documents.js
```diff
@@ -2263,12 +2264,185 @@ function writeAllDocumentsImpl(options) {
...
+          }
+        }
+        writeBatchArray.push(record);
+      }
+
+  if (writeBatchArray.length !== 0) {
+    const batchdef = [];
+    batchdef.push({documents: writeBatchArray});
+    if (jobState.jobOptions.transform) {
+      batchdef.push({transform:jobState.jobOptions.transform});
```
ehennum, line 2393: I think a comment on a prior review iteration got lost in the revisions.  I probably resolved a comment prematurely.

The transform isn't pushed on the batch definition.  It's a parameter of the call to writeDocumentsImpl().
See the JSDoc in the source file before the writeDocuments() definition for the order of the parameters.

By the way, it will be important for a unit test to confirm that the default metadata (if supplied) is written.

ehennum (approved):
Other than the transform, this increment looks ready for writing unit tests.

Then, the other aspects of the story such as error disposition can be added.

anu3990:
Combining all the above changes in a single commit - https://github.com/anu3990/node-client-api/commit/3b0c87290e264bf87c42346043306cfe20f6808c.
The develop branch will now have only one commit - "Node 9:Operation. Node 10:Finalization"

PR #602, opened 2021-09-23
Node 16:Unit tests for Data Movement ingestion.

anu3990:


ehennum (approved):
Looks great to me.  

The only thing in the story that I missed in the unit tests were the return values from onBatchError() to control the disposition of the error.

Question:  Each time the API implementation calls an application-supplied callback, is that call wrapped in a try / catch as in:

     try { ... call ... } catch(err) { ... log error and terminate job ... }

That would be important for graceful handling of coding errors in the application's callbacks.

One other thought:  It might be good to review the Implementation section of the Ingesting Documents story just to double check for final polish.

PR #608, opened 2021-09-28
Updating documents.js and unit tests.

anu3990:


lib/documents.js
```diff
@@ -2444,7 +2444,8 @@ function onWriteAllDocs(jobState, writerId) {
               case 1: {
                 jobState.retryCount--;
                 batchdef.documents = errorDisposition;
-                writeDocumentsImpl.call(jobState.docInstance, false, [batchdef]);
+                writeDocumentsImpl.call(jobState.docInstance, false, [batchdef],
+                    onWriteAllDocs(jobState, writerId));
```
ehennum, line 2448: It makes sense that this call to writeDocumentImpl should take a callback.

However, line 2448 is executing onWriteAllDocs() immediately instead of passing a function reference.

Also, writeDocumentImpl doesn't take a 3rd parameter.  Instead, the success callback is provided by chaining a result() call.

In addition to a success callback, it would be important to provide an error callback by chaining a catch() call.

If I recall correctly, the module has other instances of this writeDocumentImpl(...).result(...).catch(...) pattern.

ehennum (approved):
Aside from a concern about the callback on writeDocumentImpl(), everything looks good to me.

lib/documents.js
anu3990: Thanks Erik.!
Added writeDocumentsImpl.call(..).result(){..}.catc(){..} in a separate function "writeDocs". The writeDocumentsImpl.call is replaced by the call to writeDocs. writeDocs is also called again in case onBatchError sends a replacement batch.

PR #610, opened 2021-09-30
Node 28 : Document the writeAll Api.

anu3990:


lib/documents.js
```diff
@@ -2235,13 +2235,22 @@ Documents.prototype.suggest = function suggestDocuments() {
...
  * @method documents#writeAll
  * @since 2.8.0
- * @param {string} options - Javascript literal object that configures the write operation.
- * @returns {PassThrough stream} - a stream.Writable in object mode that receives document descriptor input from the
- * application for writing to the database.
+ * @param {function(progress, documents)} [onBatchSuccess] - Notifies other systems about batches written successfully.
+ * @param {function(progress, documents, error)} [onBatchError] - Responds to any error while writing a batch of documents.
+ * @param {function(summary)} [onCompletion] - Provides a summary of the results.
+ * @param {Integer} [batchSize] - The number of documents in each batch.
+ * @param {object[]} [concurrentRequests] - JavaScript object literal that controls the maximum number of concurrent
```
ehennum, line 2247: Given that the parameter accepts only one object, would the correct type be object instead of object[]?

The same concern applies to defaultMetadata

lib/documents.js
```diff
@@ -2235,13 +2235,22 @@ Documents.prototype.suggest = function suggestDocuments() {
...
+ *  requests that can be pending at the same time. Valid keys are "multipleOf" and "multiplier".
+ * @param {object[]} [defaultMetadata] - the metadata values applied to each document including collections,
+ *  permissions, properties, quality, and metadataValues.
+ * @param {object[]} [transform] - transformSpecification applied to each document.
```
ehennum, line 2251: Should this have the same JSDoc as writeDocuments():  @param {string|mixed[]}  [transform]

lib/documents.js
```diff
@@ -2235,13 +2235,22 @@ Documents.prototype.suggest = function suggestDocuments() {
...
+ * @returns {PassThrough|stream} - a stream.Writable in object mode that receives document descriptor input from the
```
ehennum, line 2252: The type should be stream.Writable to align with the comment.

Internally, the current implementation uses the stream.PassThrough subclass of stream.Writable, but the user doesn't need to know and that could change to a different subclass of stream.Writable at the convenience of the maintainers of the Node.js API.

ehennum (approved):
Thanks for the guidance to the application developer.

I've added a few suggestions.

Issue #553, opened 2020-07-07
update nodejs dependencies, CHANGELOG.md and release notes.


ehennum:
Including the minimum Node.js version

Issue #601, opened 2021-09-07
regenerate Optic expression functions


ehennum:
pick up the latest changes

ehennum:
As in Java, 
Added sql.bucket() and cts.columnRangeQuery()
Fixed the data types and documentation on some existing methods.

ehennum:
Should cause no regressions.

PR #613, opened 2021-10-07
#553 - Changelog update.

anu3990:


georgeajit (changes requested):
I think reference to "(Internal Jira Story)" can be removed.
For #601 Capitalize "r" in "regenerate".

ehennum (approved):

PR #616, opened 2021-10-12
#615 - Include an example for the writeAll api.

anu3990:
https://github.com/marklogic/node-client-api/issues/615

### DIFF
--- a/examples/writeAll-documents.js
+++ b/examples/writeAll-documents.js
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2021 MarkLogic Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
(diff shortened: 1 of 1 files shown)


examples/writeAll-documents.js
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+let exutil = require('./example-util.js');
+const marklogic = exutil.require('marklogic');
+const dbWriter = marklogic.createDatabaseClient(exutil.restWriterConnection);
+let uris = [];
```
ehennum, line 20: It looks like the uris array isn't used after being populated. 

Could the uris array be removed from the example to simplify and focus attention on the readable?

examples/writeAll-documents.js
```diff
@@ -0,0 +1,44 @@
...
+const Stream = require('stream');
+let readable = new Stream.Readable({objectMode: true});
```
ehennum, line 22: Could readable be a const?  The variable that refers to the stream object never changes (though the stream object changes).

examples/writeAll-documents.js
```diff
@@ -0,0 +1,44 @@
...
+}
+readable.push(null);
+
+readable.pipe(dbWriter.documents.writeAll({
+
+    onCompletion: ((summary) => {
+        console.log(summary.docsWrittenSuccessfully+' documents were written successfully.');
+        console.log(summary.docsFailedToBeWritten+' documents were not written.');
+        console.log('Time taken was '+summary.timeElapsed+' milliseconds.');
+    })
```
ehennum, line 43: Would it be good to show a failure listener as a best practice?

PR #623, opened 2021-11-05
DRAFT - Collect documents Uri

anu3990:


lib/documents.js
```diff
@@ -2472,6 +2472,95 @@ function writeDocs(jobState, val, batchdef, writeBatchArray, writerId){
...
+  requestOptions.path = mlutil.databaseParam(connectionParams, endpoint);
+  let operation = new Operation(
+      'read uris', this.client, requestOptions, 'empty', 'single'
+  );
+
+  operation.requestBody = searchBody;
+  operation.validStatusCodes  = [200, 204, 404];
+  operation.inlineAsDocument  = false;
+
+  requester.startRequest(operation).result(jobState, onQueryAllDocs)
```
ehennum, line 2553: I believe this line would need to execute the call() method on the function reference to set the this object, as in:

requester.startRequest(operation).result.call(jobState, onQueryAllDocs)

ehennum (commented):
This draft is headed in the right direction.

I think the next step is to revise to

- add concurrent URI collectors in an approach similar to the concurrent writers for writeAll()
- make the initialization and collection requests with inputs and outputs similar to the Java API

lib/documents.js
```diff
@@ -2472,6 +2473,219 @@ function writeDocs(jobState, val, batchdef, writeBatchArray, writerId){
...
+      if(i<arr.length-1){
+        const readable = stream.Readable();
+        arr.subarray(i,arr.length-1).forEach(item => readable.push(item));
+        readable.push(null);
+        jobState.stream.write(readable);
+      }
+    }
+    else {
+      jobState.stream.write(output);
+    }
```
ehennum, line 2581: If the output is null, I believe that the callback should write the null to the stream only if the forest requester is the last active requester. 

That is, my understanding is that Node.js streams end as soon as a null is written to the stream.  So, the null shouldn't be written if other forest requesters are still active.

test-basic/documents-data-movement-queryAll.js
```diff
@@ -0,0 +1,106 @@
...
+            dbWriter.documents.queryAll(q.where(query));
+        } catch(err){
+            err.toString().should.equal('Error: Query needs to be a cts query.');
+            done();
+        }
+    });
+
+    it('should queryAll documents with empty options', function(done){
+
+        dbWriter.documents.queryAll(q.where(query))
```
ehennum, line 71: Should it be possible to pass in the query directly?  Instead of having to wrap it with q.where()?

ehennum (approved):
Substantial progress!  I've made some comments for consideration.

anu3990:
Merging after adding all the changes and combining all commits into one.

Issue #622, opened 2021-10-27
Collecting Document URIs - queryAll api


anu3990:
The Node.js API documents object adds a queryAll() function equivalent to the DMSDK QueryBatcher with the following signature:

queryAll(query, options)

georgeajit:
Tests are available.

PR #624, opened 2021-11-19
Unit tests for queryAll api.

anu3990:
#622 - Unit tests for queryAll api.


ehennum (approved):
Thanks for providing this capability.

anu3990:
consistentSnapshotTimestamp will be added to summary if consistentSnapshot is set to true.

Issue #620, opened 2021-10-20
Exporting Documents - readAll api


anu3990:
The Node.js API documents object adds a readAll() function equivalent to the DMSDK ExportListener with the following signature:
readAll(options)

georgeajit:
Tests available.

PR #628, opened 2021-12-07
readAll api - Implementation and Unit tests for batchSize, onCompletion, concurrentRequests, inputKind and outputStreamType

anu3990:
https://github.com/marklogic/node-client-api/issues/620 - readAll api - Implementation and Unit tests for batchSize, onCompletion, concurrentRequests, inputKind and outputStreamType

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -23,6 +23,7 @@ var pathModule = require('path');
 var fs         = require('fs');
 const stream = require('stream');
 const bldrbase = require('./plan-builder-base.js');
+const duplexify = require('duplexify');
 
 /** @ignore */
 function addDocumentUri(documents, document) {
@@ -2292,7 +2293,7 @@ function writeAllDocumentsImpl(jobOptions) {
      const concurrentRequests = jobState.jobOptions.concurrentRequests;
      if (concurrentRequests.multipleOf) {
        if (!new Set(['forests', 'hosts']).has(String(concurrentRequests.multipleOf))) {
-         throw new Error('Invalid value for onCompletion.multipleOf. Value must be forests or hosts.');
+         throw new Error('Invalid value for multipleOf. Value must be forests or hosts.');
        }
      } else {
        concurrentRequests.multipleOf = 'forests';
@@ -2355,7 +2356,7 @@ function onWriteAllInit(output) {
           new Set(output.map((forest) => forest.host)).size;
       break;
     default: {
-      const err =  new Error('Invalid value for onCompletion.multipleOf. Value must be forests or hosts.');
+      const err =  new Error('Invalid value for multipleOf. Value must be forests or hosts.');
       jobState.error = err;
       jobState.stream.emit('error', err);
     }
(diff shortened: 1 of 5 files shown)


lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+  }
+  else {
+    jobState.jobOptions.concurrentRequests = {multipleOf:'forests', multiplier:8};
+  }
+  if(jobState.jobOptions.batchSize){
+    if(jobState.jobOptions.inputKind === 'array'){
+      throw new Error('batchSize not expected when inputKind is array.');
+    }
+    if(jobState.jobOptions.batchSize <=0){
+      jobState.jobOptions.batchSize = 250;
```
ehennum, line 2809: Would it be better to throw an error if the request supplies an invalid batch size so the application developer can fix the input?

lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+    }
+  } else {
+    jobState.jobOptions.batchSize = 250;
+  }
+
+  requester.startRequest(operation).result(mlutil.callbackOn(jobState, onReadAllInit))
+      .catch((err) => operation.stream.emit('error', err));
```
ehennum, line 2816: If I recall correctly, the operation will use a promise instead of a stream if the request takes a callback.  

If operation.stream is undefined, I suspect that the attempt to use operation.stream would throw a silent error.

During development of the API, logging might reveal whether the forest info request is succeeding.

Should the error be fired on the input stream for the document URIs?

lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+  if(jobState.error){
+    finishReader(jobState);
+    return;
+  }
+  const readBatchArray = [];
+  const batchSize = jobState.jobOptions.batchSize;
+  jobState.lastBatch = false;
+  jobState.retryCount = 100;
+
+  a:
```
ehennum, line 2852: Should this function test jobState.jobOptions.inputKind and 

* if string, iterate to batchsize
* if array, use the array as is

lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+      for (let i = 0; i < batchSize; i++) {
+        const record = jobState.inputStream.read();
```
ehennum, line 2854: It might be more readable for future maintenance to rename the record constant to uri because the string represents a document uri.

lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+    const batchdef = {uris: readBatchArray};
+    if(jobState.jobOptions.transform){
+      batchdef.transform = jobState.jobOptions.transform;
+    }
+    readDocs(jobState, true, batchdef,readBatchArray, readerId);
+  }
+}
+
+function readDocs(jobState, val, batchdef, readBatchArray, readerId){
+  readDocumentsImpl.call(jobState.docInstance, val, [batchdef])
```
ehennum, line 2880: As I read the source code for readDocumentsImpl(), the 3rd argument should be either a simple array of uris or an object with a uris property.

As I read onReadAllDocs(), the batchdef parameter is an object with a uris property.

So, I suspect the call to readDocumentsImpl() should pass the batchdef without wrapping it in an array.

Also, when producing chunked output, it will be necessary to get a chunked stream from readDocumentsImpl()
The object case is much more important, so it is best to focus on that first.  It might even make sense to defer support for chunked streams to a future release.

lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+      .result((output) => {
+        jobState.docsReadSuccessfully+= readBatchArray.length;
```
ehennum, line 2882: Should the result handler write each output document to the output stream for the job state?

lib/documents.js
```diff
@@ -2716,6 +2716,253 @@ function urisOutputTransform(headers, data) {
...
+            break;
+          }
+        }
+      });
+}
+
+function finishReader(jobState) {
+  jobState.requesterCount--;
+  if(jobState.requesterCount === 0) {
+    console.log('inside finishReader '+jobState.docsReadSuccessfully)
```
ehennum, line 2951: The console.log() should be removed after development of the API.

I would expect that the finalization code should end the output stream.

test-basic/documents-data-movement-readAll.js
```diff
@@ -0,0 +1,71 @@
...
+            .result(function(response){
+                done();
+            })
+            .catch(err=> done(err))
+            .catch(done);
+    }));
+
+    it('should readAll documents with empty options', function(done){
+
+        uris.pipe(dbWriter.documents.readAll()).pipe(fs.createWriteStream('file.txt'));
```
ehennum, line 59: I suspect that a readable stream can only be read once.

If so, it may be necessary to construct the urisList from the Stream.Readable separately for the initialization and the test.

ehennum (changes requested):
This reflects demonstrates progress and, I suspect, is close to a working implementation.

I hope the comments help to get closer.

lib/documents.js
```diff
@@ -2716,6 +2717,284 @@ function urisOutputTransform(headers, data) {
...
+  if(jobState.error){
+    finishReader(jobState);
+    return;
+  }
+  readDocumentsImpl.call(jobState.docInstance, val, [batchdef])
+      .result((output) => {
+        jobState.docsReadSuccessfully+= readBatchArray.length;
+        output.forEach(item => {
+          if(jobState.readableObjectMode === 'chunked'){
+            jobState.outputStream.write(JSON.stringify(item));
```
ehennum, line 2908: As discussed in email, it would be much better in the chunking case to 

- read the response to the request as a stream
- listen for data events on the response stream and write data events to the job output stream
- listen for error events on the response stream with error disposition logic
- listen for the end event to determine when to fire the next request

Avoiding stringify is important for performance reasons, which is at a premium for bulk data movement.

test-basic/documents-data-movement-readAll.js
```diff
@@ -0,0 +1,236 @@
...
+                if(err){
+                    done(err);
+                }
+                arr.forEach(item=> result.add(item.uri)
+                );
+                checkResult(done);
+            });
+    });
+
+    it('should readAll documents with inputKind option as array', function(done){
```
ehennum, line 110: A good test might be to 

- construct a readable stream in object mode
- iterate over urislist in increments of 10
- on each iteration, call slice(i, i + 10) to construct an array of 10
- add each array of 10 as a single item to the readable stream
- pipe the readable stream to readAll()

test-basic/documents-data-movement-readAll.js
```diff
@@ -0,0 +1,236 @@
...
+
+    it('should readAll documents with outputStreamType option as chunked', function(done){
+        streamToArray(uriStream.pipe(dbWriter.documents.readAll({
+                outputStreamType: 'chunked'
+            })),
+            function(err, arr ) {
+                if(err){
+                    done(err);
+                }
+                arr.forEach(item=> result.add(JSON.parse(item).uri));
```
ehennum, line 158: For a chunked stream, the chunks might not be parseable individually.

ehennum (approved):
This looks great.  I'd suggest 

- Fixing up the array input unit test
- Working on chunking after merging this work

anu3990:
Combining all the code changes in 4 commits into one commit.

PR #630, opened 2021-12-20
readAll api - Implementation for chunked outputStreamType and unit tests for transform, categories, consistentSnapshot and onBatchError.

anu3990:
https://github.com/marklogic/node-client-api/issues/620 - Implementation for chunked outputStreamType and unit tests for transform, categories, consistentSnapshot and onBatchError.

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -2811,6 +2811,12 @@ function readAllDocumentsImpl(jobOptions) {
     jobState.jobOptions.batchSize = 250;
   }
 
+  if(jobState.jobOptions.outputStreamType){
+    if(jobState.jobOptions.outputStreamType === 'chunked' && jobState.jobOptions.categories){
+      throw new Error('categories not expected when outputStreamType is chunked.');
+    }
+  }
+
   if(jobState.jobOptions.consistentSnapshot && typeof jobState.jobOptions.consistentSnapshot !== 'boolean'){
     throw new Error('consistentSnapshot needs to be a boolean.');
   }
@@ -2838,7 +2844,20 @@ function onReadAllInit(output) {
       break;
   }
   jobState.startTime = Date.now();
-  for (let i = 0; i < maxRequesters; i++) {
+  jobState.maxRequesters = maxRequesters;
+
+  if(jobState.jobOptions.consistentSnapshot){
+    jobState.requesterCount++;
+    onReadAllDocs(jobState,0);
+  }
+  else {
+    spinReaderThreads(jobState,0);
+  }
+}
+
+function spinReaderThreads(jobState, readerId){
+
+  for(let i=readerId;i<jobState.maxRequesters; i++){
     jobState.requesterCount++;
     onReadAllDocs(jobState,i);
   }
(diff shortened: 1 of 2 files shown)


lib/documents.js
```diff
@@ -2904,78 +2922,91 @@ function readDocs(jobState, val, batchdef, readBatchArray, readerId){
...
+          jobState.outputStream.write(item);
+        })
+        .on('end', function(){
           onReadAllDocs(jobState, readerId);
-        }
-      })
+        });
+  }
+  else {
+    readDocumentsImplRequest.result((output) => {
```
ehennum, line 2939: In object mode, the timestamp would first become available in this callback and thus other requesters could be spun up.

lib/documents.js
```diff
@@ -2904,78 +2922,91 @@ function readDocs(jobState, val, batchdef, readBatchArray, readerId){
...
+      jobState.docsReadSuccessfully+= readBatchArray.length;
```
ehennum, line 2940: I see belatedly that, in chunked mode, the Node.js API cannot report the number of documents read successfully or failed.

Maybe that's something to document?

test-basic/documents-data-movement-readAll.js
```diff
@@ -216,19 +235,179 @@ describe('data movement readAll', function() {
...
+                    'metadataKey2': 'metadataValue 2'
+                },
+                collections: 'collection1',
+                permissions: [
+                    {'role-name': 'app-user', capabilities: ['read']},
+                    {'role-name': 'app-builder', capabilities: ['read', 'update']}
+                ],
+                quality: 1
+            }
+        }));
```
ehennum, line 272: If I recall correctly, Mocha has a before task that could do the setup separately from the it test, which would be better than using a timeout to execute the test after the setup.  

Or, the test could move into an on completion callback for the setup similar to some of the other tests to execute in a result callback.

(The concern is that the timeout could be fragile.)

test-basic/documents-data-movement-readAll.js
```diff
@@ -216,19 +235,179 @@ describe('data movement readAll', function() {
...
+            onCompletion: ((summary) => {
+                summary.docsReadSuccessfully.should.be.equal(10);
+                summary.docsFailedToBeRead.should.be.equal(0);
+                summary.timeElapsed.should.be.greaterThanOrEqual(0);
+                done();
+            })
+        }));
+    });
+
+    it('should readAll documents with consistentSnapshot option as true', function(done){
```
ehennum, line 375: A good test for a consistent snapshot would be to update a document after the first document is read but before the modified document is read.

The output should have the old content for the modified document because the timestamp of the modification is after the consistent timestamp.

ehennum (approved):
Looks very close with a good level of testing.

lib/documents.js
```diff
@@ -2838,7 +2844,22 @@ function onReadAllInit(output) {
...
   }
   jobState.startTime = Date.now();
-  for (let i = 0; i < maxRequesters; i++) {
+
+  if(jobState.jobOptions.consistentSnapshot){
+    jobState.consistentSnapshotTimestamp = new mlutil.Timestamp();
+    jobState.requesterCount++;
+    let firstRequestCompleted = onReadAllDocs(jobState,0);
+    if(firstRequestCompleted){
+      spinReaderThreads(jobState,1, maxRequesters);
```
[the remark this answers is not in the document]
anu3990: Hi Erik, this is one of the ways to make the code synchronous. The variable "firstRequestCompleted" will be true only after 2851 is executed, so line 2853 is always executed after the first request. I had put loggers in the function readDocumentsImpl for timestamp values in each request and only the first value came as null. Rest all requests had values in them.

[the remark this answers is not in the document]
ehennum: Hi, Anushree:  onReadAllDocs() will certainly have executed.

But, executing onReadAllDocs() only starts the first request.  The response from the first request won't come back until later when the object or chunked mode response callback executes.

Because the timestamp is in the response, the timestamp won't be available until the response comes back.

If onReadAllDocs() returns a value, the firstRequestCompleted variable will evaluate to true, but that doesn't guarantee the timestamp is available.

Am I missing something?

test-basic/documents-data-movement-readAll.js
anu3990: I will work on this.

anu3990:
Combining all commits into one commit before merge.

PR #636, opened 2022-01-17
Adding JS Docs and onInitialTimestamp for readAll api.

anu3990:


ehennum (approved):
Thanks for catching that

PR #642, opened 2022-02-01
Add a category option for reading only the contents of document(s)

anu3990:


ehennum (approved):
My only concern is that I'm not entirely sure from the test whether the response is providing a document descriptor (which would be an issue) or just the document content (which is the goal).

I'm approving with the assumption that, if I understood the test, I would realize that the response contains only the content.

anu3990:
Thanks for taking a look @ehennum .!
Yes, the test "should queryToReadAll documents with categories options rawContent" confirms that each json returned has only one key (keyCount variable in the test) and that each key contains the string "key" in it. Currently the returned jsons look like - '{ 'key 165': 'value 165' }'. According to the before block, the list of uris we are quering 'categoriesUrisList' has the word 'key' only in the content. If the entire document descriptor was returned, the returned value will have the following structure -
{
  uri: '/test/dataMovement/requests/categories/365.json',
  category: 'content',
  format: 'json',
  contentType: 'application/json',
  contentLength: '23',
  content: { 'key 365': 'value 365' }
}
and then the test will fail with the exception - Uncaught AssertionError: expected 'uri' to contain 'key' (line 524)

ehennum:
Great -- that completely addresses my hesitation, thanks for confirming.

PR #650, opened 2022-02-12
Draft: transformAll

anu3990:


lib/documents.js
```diff
@@ -3338,6 +3338,280 @@ function copyConstraints(suggestConstraints, searchConstraints) {
...
+  if(jobState.jobOptions.transformStrategy){
+    endpoint = endpoint+'&result='+String(jobState.jobOptions.transformStrategy);
+  }
+
+  let connectionParams = jobState.docInstance.client.getConnectionParams();
+  let requestOptions = mlutil.copyProperties(connectionParams);
+  requestOptions.method = 'POST';
+  requestOptions.headers = {
+    'Accept':       'text/uri-list',
+    'Content-Type': 'application/json'
```
ehennum, line 3448: I believe the /v1/internal/apply-transform request should set both the Content-Type header and the Accept header to the text/uri-list mime type.

The request sends a list of URIs to transform and receives a list of URIs that were transformed.

lib/documents.js
```diff
@@ -3338,6 +3338,280 @@ function copyConstraints(suggestConstraints, searchConstraints) {
...
+  };
+  requestOptions.path = mlutil.databaseParam(connectionParams, jobState.endpoint);
+
+  let operation = new Operation(
+      'transform docs', jobState.docInstance.client, requestOptions, 'multipart', 'single'
```
ehennum, line 3453: I believe both the request payload and response payloads should be 'single'

The URI list sent for transformation should be a single payload instead of sending multiple HTTP body payloads as parts.

lib/documents.js
```diff
@@ -3338,6 +3338,280 @@ function copyConstraints(suggestConstraints, searchConstraints) {
...
+
+  requestOptions.method = 'GET';
+  requestOptions.headers = {
+    'Accept':       'application/json',
+  };
+
+  requestOptions.path = mlutil.databaseParam(connectionParams, path);
+  let operation = new Operation(
+      'read forestInfo', this.client, requestOptions, 'empty', 'single'
+  );
```
ehennum, line 3360: Could any of the setup for getting forest information be refactored into a common function to reduce duplication of code and thus make maintenance easier?

lib/documents.js
```diff
@@ -3338,6 +3338,280 @@ function copyConstraints(suggestConstraints, searchConstraints) {
...
+            break;
+          }
+          default: {
+            jobState.docsFailedToBeTransformed += transformBatchArray.length;
+            jobState.error = err;
+            jobState.stream.emit('error', err);
+            finishTransformAll(jobState);
+          }
+        }
+      });
```
ehennum, line 3595: Could any of the retry logic be refactored into a common function called from multiple error handlers and thus make the implementation easier to maintain?

test-basic/documents-data-movement-transformAll.js
```diff
@@ -0,0 +1,84 @@
...
+let xqyTransformPath = './test-basic/data/flagTransform.xqy';
+let readable = new Stream.Readable({objectMode: true});
+let transformStream = new Stream.PassThrough({objectMode: true});
+let uris = [];
+
+describe('data movement transformAll', function() {
+
+    before(function (done) {
+        restAdminDB.config.transforms.write(xqyTransformName, 'xquery', fs.createReadStream(xqyTransformPath))
+            .result(function(response){})
```
ehennum, line 42: I believe the call to result() can be empty to specify the Promise success callback in a chained then() call. 

(In fact, I'm not sure the call to result() will work with a function that doesn't return anything.)

test-basic/documents-data-movement-transformAll.js
```diff
@@ -0,0 +1,84 @@
...
+let uriStream = new Stream.PassThrough({objectMode: true});
+let urisList = [];
+let result = new Set();
+let summaryValue = null;
+let categoriesUrisList = [];
+let xqyTransformName = 'flagParam';
+let xqyTransformPath = './test-basic/data/flagTransform.xqy';
+let readable = new Stream.Readable({objectMode: true});
+let transformStream = new Stream.PassThrough({objectMode: true});
+let uris = [];
```
ehennum, line 36: If these variables are only needed inside the describe(), it would be better to declare them inside the describe()?

Minimizing scope is a good habit to cultivate even in tests, because it can minimize unexpected interactions.

test-basic/documents-data-movement-transformAll.js
```diff
@@ -0,0 +1,84 @@
...
+
+describe('data movement transformAll', function() {
+
+    before(function (done) {
+        restAdminDB.config.transforms.write(xqyTransformName, 'xquery', fs.createReadStream(xqyTransformPath))
+            .result(function(response){})
+            .then(function(documents){
```
ehennum, line 43: The documents parameter implies that the config.transforms.write() request returns documents.  Does it?   

If not, a less specific name like output would make the code easier to read by future maintainers without confusion.

ehennum (commented):
This looks very far along.

I've made some incidental comments.  One possible resolution of the refactoring comments is to log an issue to address in the future.

Hoping that's useful...

lib/documents.js
anu3990: Created a new issue to refactor the code - https://github.com/marklogic/node-client-api/issues/652

PR #662, opened 2022-03-04
Implementation and unit tests for transformAll api.

anu3990:


ehennum (approved):
I had a few incidental remarks, but the capability looks correct and tested to me.

PR #663, opened 2022-03-10
NODE-58 Add unit test(s) for op:bucket-group in Node API

AndreiPanti:


test-basic/plan-documents.js
```diff
@@ -50,7 +50,88 @@ describe('documents', function() {
...
+
+                ])
+                    .groupToArrays(
+                        p.bucketGroup("rBucket", p.col("r"), [0, 8]),
+                        p.count("numRows")
+                    )
+            )
+                .then(function (response) {
+                    const output = getResults(response);
+                    should(output[0]['rBucket']['value'].length).equal(1);
```
anu3990, line 127: This test will throw assertion exception, since the number of bucket created will be 2 ([
  { rBucket_bucket: 1, numRows: 7 },
  { rBucket_bucket: 2, numRows: 1 }
] ). The upper limit is not included

anu3990 (changes requested):

test-basic/plan-documents.js
AndreiPanti: I tested and the number of buckets is 1 ( for this example, where the boundaries are [0,8].

image

anu3990: Hi @AndreiPanti , please upgrade the server to 10.0.9

anu3990 (approved):

Issue #644, opened 2022-02-02
Reprocessing documents - transformAll(options)


anu3990:
The Node.js API documents object adds a transformAll() function equivalent to the DMSDK ApplyTransformListener

PR #664, opened 2022-03-10
Implementation, unit tests and JS docs for queryToTransformAll api.

anu3990:
https://github.com/marklogic/node-client-api/issues/644 - Implementation, unit tests and JS docs for queryToTransformAll api.

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -2600,8 +2600,8 @@ function onQueryUrisRequest(output) {
   jobState.requesterCount = output.forests.length;
   jobState.startTime = Date.now();
   jobState.endpoint = endpoint;
-  if(jobState.jobOptions.queryToReadAll){
-    jobState.jobOptions.readAllJobOptions.output = output;
+  if(jobState.jobOptions.queryToOptions){
+    jobState.jobOptions.queryToOptions.output = output;
   }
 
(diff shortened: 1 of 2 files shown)


georgeajit (changes requested):
Reviewd the JSDoc content and requested changes in the zoom meeting.

lib/documents.js
```diff
@@ -3661,6 +3660,112 @@ function finishTransformAll(jobState) {
...
+    transformAllJobOptions.onBatchSuccess = jobOptions.onBatchSuccess;
+  }
+
+  queryAllJobOptions.queryToOptions = transformAllJobOptions;
+
+  const docInstance = this;
+  let queryAllStream = queryAllDocumentsImpl.call(docInstance, query, queryAllJobOptions);
+  let pipelined = false;
+
+  queryAllStream.on('data', function(item){
```
ehennum, line 3755: I'm still wondering whether there is an idiomatic approach in Node.js that can pause the transformAllStream until the queryAllStream is ready to send data so the streams can be piped initially instead of waiting for the first data event to pipe.

But, I think there's already an issue to investigate that.  This would just be another potential case for refactoring to add  to that issue.

ehennum (approved):
Looks good.

anu3990:
@georgeajit completed the JS doc changes as discussed yesterday.

anu3990:
Merging this pull request after squashing all the commits into one.

PR #695, opened 2022-11-23
NODE-83: Update "with fragment id column" test in plan-triples.js

mariusmarklogic:


test-basic/plan-triples.js
```diff
@@ -175,11 +175,14 @@ describe('triples', function() {
         should(output[0].datastore.value).equal('/datastore/id#A');
         should(output[0].title.value).equal('The A datastore');
         should(output[0].sourceDocCheck.value).equal(true);
-        should.not.exist(output[0].sourceDocId);
+        should.exist(output[0].sourceDocId);
+        should(output[1].datastore.value).equal('/datastore/id#B');
```
anu3990, line 179: We don't need line 179. This check - "should(output[1].datastore.value).equal('/datastore/id#B');" is already covered on line 181.

anu3990 (changes requested):

anu3990 (approved):

PR #698, opened 2022-12-10
NODE-90: FromDocUris implementation in node-client-api

mariusmarklogic:


test-basic/optic-fromDocUris.js
```diff
@@ -0,0 +1,93 @@
...
+var testconfig = require('../etc/test-config.js');
+const execPlan = pbb.execPlan;
+const getResults = pbb.getResults;
+var db = marklogic.createDatabaseClient(testconfig.restWriterConnection);
+var q = marklogic.queryBuilder;
+const Stream = require('stream');
+let result = new Set();
+
+describe('fromDocUris', function() {
+    this.timeout(15000);
```
rjrudin, line 18: @mariusmarklogic  @anu3990  Any time there's a timeout/sleep in a test, we should have a comment to explain why. We want tests to run as quickly as possible, so we only want a timeout if we really need it - and then we need that comment to explain why we need it.

README.md
```diff
@@ -26,6 +26,10 @@ using [npm](https://www.npmjs.com/package/marklogic):
 ```
 npm install marklogic --save
 ```
+For Windows OS please use the below for Node Client 2.9.1:
```
rjrudin, line 29: Shouldn't this addition be coming from a separate PR?

mariusmarklogic: yes, but i don't remember why this isn't coming then :)

anu3990: Hi Marius, this is not needed. Please remove this change.

test-basic/optic-fromDocUris.js
anu3990: Hi Rob, this is a common timeout for all the tests in this file. This is primarily for Jenkins. Mocha has a timeout of 2 seconds. On jenkins we use mocha to run the tests. Due to the sync/async path Node follows, it sometimes takes more than 2 seconds to complete. So this is a way to override the mocha auto-timeout.

test-basic/optic-fromDocUris.js
```diff
@@ -0,0 +1,93 @@
...
+            .then(function(response) {
+                const rows = response.rows;
+                rows.forEach(item => result.add(item.uri.value));
+                checkResult(done);
+            })
+            .catch(err=> done(err));
+    });
+
+    it('with query-2', function(done){
+        // similar to test-basic/rows.js line 160
```
anu3990, line 60: @mariusmarklogic  please remove this and other commented out lines in this file.

anu3990 (changes requested):

PR #704, opened 2023-01-10
NODE-113 : Connect to MarkLogic instance with base path

anu3990:


test-basic/basePath-test.js
```diff
@@ -0,0 +1,92 @@
...
+            .catch(err=>
+            {
+                assert(err.toString().includes('basePath invalid/v1/documents'));
+                done();
+            });
+    });
+
+    it('basePath with multiple starting and trailing slashes', function(done){
+        testconfig.restWriterConnectionWithBasePath.basePath = '//invalid//';
+        const dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnectionWithBasePath);
```
anu3990, line 78: This test currently fails because we remove only one slash before and after the word "invalid" and hence the basePath it tries to connect is - "/invalid//v1/documents"

BillFarber (approved):

test-basic/basePath-test.js
anu3990: Updated this test as per the discussion in this morning's meeting. Thanks.!

lib/mlutil.js
```diff
@@ -155,6 +155,12 @@ function endpointTransform(transform) {
 function databaseParam(connectionParams, endpoint, separator) {
   var database = connectionParams.database;
   if (database == null) {
+    if(connectionParams.basePath) {
```
rjrudin, line 158: This is surprising - is `databaseParam` really guaranteed to be called for every single call to the REST API?? 

I know there are multiple REST API endpoints that do not accept "database" as a parameter - for instance:

- https://docs.marklogic.com/REST/GET/v1/rest-apis/[name]
- https://docs.marklogic.com/REST/GET/v1/config/resources
- https://docs.marklogic.com/REST/GET/v1/config/indexes

If `databaseParam` isn't called for when those endpoints are invoked - and it seems surprising that it would be called, since those endpoints don't support the "database" parameter - then the basePath won't be applied, and the call will fail.

lib/mlutil.js
```diff
@@ -155,6 +155,12 @@ function endpointTransform(transform) {
...
+      let basePath = connectionParams.basePath.toString().startsWith('/') ? connectionParams.basePath.toString().substring(1):connectionParams.basePath;
```
rjrudin, line 159: This code will be more readable if you first do:

    const basePath = connectionParams.basePath.toString();

And then to make it clear that the user's original base path isn't going to be altered (which makes code easier to understand), I would do:

    let fixedBasePath = basePath.startsWith("/") ? basePath.substring(1) : basePath;
    fixedBasePath = fixedBasedPath.endsWith("/") ? fixedBasedPath.substring(0, fixedBasePath.length - 1) : fixedBasePath;
    endpoint = fixedBasedPath + endpoint;

test-basic/basePath-test.js
```diff
@@ -0,0 +1,91 @@
...
+};
+
+describe('basePath tests', function() {
+
+    it('basePath without slash(s)', function(done){
+        testconfig.restWriterConnectionWithBasePath.basePath = 'invalid';
+        const dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnectionWithBasePath);
+        dbWriter.documents.write(writeObject)
+            .result(function(response){})
+            .catch(err=>
```
rjrudin, line 33: I don't think this is really testing basePath sufficiently here - it's forcing an error, and it's instead verifying that the error message has an expected string in it. That doesn't tell us that the basePath is being constructed correctly, it's just telling us that an error message indicates that the basePath might have been constructed correctly. 

I think it would be better if this was more of a unit test that doesn't make a call to ML, but is instead like `NewBaseUrlTest` in the Java Client - i.e. it should test some method in the client that is responsible for constructing a URL based on a user's request. And then the tests verify that the correct URL is returned. No need to call MarkLogic, we just want to make sure the URL is built correctly.

rjrudin (changes requested):

PR #706, opened 2023-01-12
NODE-112: Run Optic plan using unnest functions

mariusmarklogic:


test-basic/unnestTest.js
```diff
@@ -0,0 +1,188 @@
...
+                { department: "Marketing", teamMembers: null }
+            ]
+        };
+
+        db.documents.write({
+            uri: '/optic/test/office1.json',
+            contentType: 'application/json',
+            content: doc
+        }).result(function() {
+            done();
```
rjrudin, line 41: Just curious, is it necessary to call `done()` here? (I'm not familiar with tests in this project, so I'm asking so I know if this needs to be done when I try to write a test)

test-basic/unnestTest.js
```diff
@@ -0,0 +1,188 @@
...
+        }, function(error) {
+            done(error);
+        });
+    });
+    after((function(done){
+        db.documents.remove(['/optic/test/office1.json'])
```
rjrudin, line 47: Best practice for tests is to delete data before a test starts, ensuring a clean database, and then that allows for not deleting data at the end. Which makes debugging a lot easier too, because you can see what data is in the database in case the test fails. 

Does this have to be deleted? I'm wondering if the best practice of "delete all data, except documents that are expected to always exist" can be implemented instead. I can guarantee that this makes the test less useful too - i.e. it's easy for me to explain how unnest works by running the test and then showing someone what data is in the database.

test-basic/unnestTest.js
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright (c) 2021 MarkLogic Corporation
+ *
```
anu3990, line 3: Since this is a new file please change the year to 2023.

anu3990 (changes requested):

test-basic/unnestTest.js
anu3990: Hi Rob, yes done() is needed towards the end else the test hangs saying - "Timeout  exceeded. For async tests and hooks, ensure "done()" is called;". done() is needed individually for all the blocks including before, beforeEach, after, afterEach and the tests.

anu3990:
@mariusmarklogic just for verification purposes, can you please add few console.log in the unnestInner and unnestLeftOuter functions in plan-builder-generated.js and check if they are invoked while running the above tests (No need to commit the console.logs, you can verify that in your local workspace). Thanks.!

mariusmarklogic:
I put the functionality for deleting the file using the test before all the other, i put some console logs to be sure that unnest functions are called from plan generated and i changed 2021 to 2023.

test-basic/unnestTest.js
```diff
@@ -0,0 +1,188 @@
...
+const op = marklogic.planBuilder;
+
+const TEAM_MEMBER_NAME_COLUMN = "teamMemberName";
+
+
+describe('test unnest function', function () {
+
+    before(function (done) {
+
+        db.documents.remove(['/optic/test/office1.json'])
```
rjrudin, line 29: I like this, and what I recommend trying to do later on (not in this PR) is adding a "delete everything that ought to be deleted" function that every test uses (ideally, this can be done in one place in the test code so it doesn't have to be repeated in every test).

rjrudin (commented):
LGTM, though I defer to @anu3990  for approval.

anu3990 (approved):

PR #708, opened 2023-01-26
NODE-114 : Connect to MarkLogic Cloud instance

anu3990:


BillFarber (approved):

lib/marklogic.js
```diff
@@ -727,7 +727,7 @@ MarkLogicClient.prototype.endpointCaller =function endpointCaller(endpointDeclar
 function initClient(client, inputParams) {
   var connectionParams = {};
   var isSSL = (inputParams.ssl == null) ? false : inputParams.ssl;
-  var keys = ['host', 'port', 'database', 'user', 'password', 'authType', 'token', 'basePath'];
+  var keys = ['host', 'port', 'database', 'user', 'password', 'authType', 'token', 'basePath','apikey','tenancyUrl'];
```
rjrudin, line 730: As mentioned in our meeting, I think `host` is what a user would expect, as "tenancy URL" refers to the entire URL. 

Also, the Java Client is using "cloudApiKey" (and ml-gradle is using "mlCloudApiKey"). I think it's good to include the "cloud" prefix, as "API key" by itself is a little vague. And since "API" and "key" are two separate words, the "K" should be capitalized.

lib/marklogic.js
```diff
@@ -741,17 +741,23 @@ function initClient(client, inputParams) {
...
+  }
+  if(connectionParams.apikey){
+    connectionParams.host = connectionParams.tenancyUrl;
+    connectionParams.port = 443;
+  }
   connectionParams.lookup = prefLookup;
 
-  const authType = connectionParams.authType.toUpperCase();
+  const authType = connectionParams.authType? connectionParams.authType.toUpperCase():null;
   if (authType !== 'NONE' &&
```
rjrudin, line 756: Shouldn't this really say "If authType is digest or basic, and either user or password is null, then throw an error"? 

I ask because I got this error when I did "apiKey" instead of "apikey". And it's confusing because the error is assuming that I'm trying to do digest or basic authentication.

anu3990:
Adding changes from this morning's discussion.

lib/marklogic.js
rjrudin: I see "CLOUD" is added to the list, but I think the code would be simpler and more future-proof if it instead said "If authType is digest or basic" as opposed to maintaining a list of all the other auth types.

lib/marklogic.js
```diff
@@ -772,6 +776,11 @@ function initClient(client, inputParams) {
       );
     }
     connectionParams.auth = 'SAML token='+token;
+  } else if(authType === 'CLOUD') {
+    if(!connectionParams.apiKey) {
+      throw new Error('apiKey needed for ML cloud authentication.');
```
rjrudin, line 781: Spell out "MarkLogic" here instead of "ML" (this is something we're trying to do across all of our tools to avoid any confusion over the frequent usage of "ML" as an abbreviation for "machine learning").

lib/requester.js
```diff
@@ -55,6 +56,33 @@ function getAuthenticatorKerberos(client) {
   return client.authenticatorKerberos;
 }
 
+function getAccessToken(operation, callback){
+
+  let postData = `${encodeURI('grant_type')}=${encodeURI('apikey')}&${encodeURI('key')}=${encodeURI(operation.client.connectionParams.apiKey)}`;
```
rjrudin, line 61: Small thing - I believe `const` is generally preferred whenever possible to make it clear that something won't change, whereas `let` implies that the variable assignment may be altered later on. I recommend using `const` whenever possible - I think this is one of those things JS does better than Java, as Java requires an additional keyword of `final` whereas in JS, you just use `const` instead of `let` or `var`.

anu3990: Done.

rjrudin (approved):

PR #709, opened 2023-02-01
NODE-117 : Read and write to a database different from "Documents" on ML cloud instance does not work.

anu3990:
Bug fix

### DIFF
--- a/lib/mlutil.js
+++ b/lib/mlutil.js
@@ -153,14 +153,14 @@ function endpointTransform(transform) {
 }
 
 function databaseParam(connectionParams, endpoint, separator) {
-  var database = connectionParams.database;
+  let database = connectionParams.database;
+  if(connectionParams.basePath) {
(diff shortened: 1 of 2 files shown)


rjrudin (commented):
There should be tests that can be updated as part of this, right? Like I think unit tests directly on this function make sense, and they would verify that the expected "endpoint" is returned.

I also think it's very much worth renaming this function to capture what it actually does, which is "build a request path", as every caller of this function assigns the returned value to `requestOptions.path`. Calling it `databaseParam` obscures what it's actually doing and also hides the fact that it processes the new `basePath` option.

anu3990:
@rjrudin , currently the database name is cropped off from the error thrown due to - https://github.com/marklogic/node-client-api/blob/develop/lib/operation.js#L228. But i verified for the successful cases that the document is written to the database mentioned with the "database" parameter. For the renaming i will create a separate Jira

rjrudin (approved):

PR #711, opened 2023-02-08
Solve tests failing

mariusmarklogic:
I don't know exactly why the GitHub sees the changes like this, because in my IDE appears well.

### DIFF
--- a/test-basic-proxy/lib/positive/sessionsTest.js
+++ b/test-basic-proxy/lib/positive/sessionsTest.js
@@ -22,139 +22,138 @@ const testutil = require('../testutil');
 const Sessions = require("./sessions.js");
 
 describe('session service', function() {
-  const service = Sessions.on(testutil.makeAdminClient());
+    this.timeout(20000);
+    const service = Sessions.on(testutil.makeAdminClient());
 
-  it('field', function(done) {
-    const timestampField = "timestamped";
-    const fieldSession = service.createSession();
-    let timestampValue = null;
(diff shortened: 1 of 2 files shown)


test-basic-proxy/lib/positive/sessionsTest.js
```diff
@@ -22,139 +22,142 @@ const testutil = require('../testutil');
...
+                .then(output => {
+                    done();
+                })
+                .catch(err => {
+                    expect(err.toString()).to.equal(
+                        'Error: call to /dbf/test/sessions/beginTransactionNoSession.sjs: cannot process response with 500 status'
+                    );
+                    done();
+                }).finally(() => {
+                done();
```
anu3990, line 145: @mariusmarklogic instead of calling done() in finally block, we need to investigate why it is failing. I recommend adding a try-catch block around line 140 and see if the failure is there. Same for line line 158

anu3990 (changes requested):

test-basic-proxy/lib/positive/sessionsTest.js
mariusmarklogic: I put the finally block because the error seemed to be the exact same thing when i looked for the first time, but now I observed that it seems to be from Operation.js(227), which includes "with path: " after the error. I changed it assertion from "equal" to "contain".

anu3990:
Closing this since it was covered in the PR - https://github.com/marklogic/node-client-api/pull/717

Issue #713, opened 2023-02-09
ReferenceError: parse_params is not defined


PavlosIsaris:
In the newer NodeJS versions, when trying to `require` the library, the following error get thrown:

```
"ReferenceError: parse_params is not defined",
    "    at node_modules/www-authenticate/lib/parsers.js (/var/task/index.js:18318:18)",
    "    at __require (/var/task/index.js:9:50)",
    "    at node_modules/www-authenticate/lib/www-authenticate.js (/var/task/index.js:18463:19)",
    "    at __require (/var/task/index.js:9:50)",
    "    at node_modules/www-authenticate/index.js (/var/task/index.js:18633:28)",
    "    at __require (/var/task/index.js:9:50)",
    "    at node_modules/marklogic/lib/requester.js (/var/task/index.js:24206:33)",
    "    at __require (/var/task/index.js:9:50)",
    "    at node_modules/marklogic/lib/documents.js (/var/task/index.js:30830:21)",
    "    at __require (/var/task/index.js:9:50)"
```

This is because of the [www-authenticated](https://github.com/randymized/www-authenticate) library.
A relevant [pull request](https://github.com/randymized/www-authenticate/pull/2#issue-1569820441) that solves the issue has already been opened, but not merged yet.
Do you propose any other ways to solve it?

anu3990:
@PavlosIsaris Thank you for the fix.! Currently Node-client uses 0.6.2 version of www-authenticate. Is this fix available in the 0.6.3 version? We can upgrade to 0.6.3.

PavlosIsaris:
@anu3990 As I can see from the [www-authenticate repo](https://github.com/randymized/www-authenticate), it was marked as "archived", meaning that the library is slowly getting abandoned. So, I don't think that there will be a 0.6.3 release. But maybe the author added the fix in the 0.6.2 release. 
Anyway, it will be a good idea to abandon this library, and use another, equivalent one.

PR #715, opened 2023-02-14
NODE-111 : Upgrade the Node-Client-Api to support an upgraded NODE 18

anu3990:
Fix for -
1. NODE-111 : Upgrade the Node-Client-Api to support an upgraded NODE 18
2. Jenkins test failures.
3. https://github.com/marklogic/node-client-api/issues/713

### DIFF
--- a/lib/documents.js
+++ b/lib/documents.js
@@ -2260,6 +2260,7 @@ Documents.prototype.suggest = function suggestDocuments() {
  *  defaultMetadata and transform.
  * @method documents#writeAll
  * @since 2.8.0
+ * @param {stream} [stream] - stream containing the input.
  * @param {object} [options] - Configures the write operation.
  * @param {function(progress, documents)} [onBatchSuccess] - A callback that can notify other systems about batches
  *  written successfully.
@@ -2275,18 +2276,17 @@ Documents.prototype.suggest = function suggestDocuments() {
  * @returns {stream.Writable} - a stream.Writable in object mode that receives document descriptor input from the
  *  application for writing to the database.
  */
-Documents.prototype.writeAll = function writeAllDocuments(options) {
+Documents.prototype.writeAll = function writeAllDocuments(stream,options) {
   return writeAllDocumentsImpl.call(
-      this, options
+      this, stream,options
   );
 };
-function writeAllDocumentsImpl(jobOptions) {
+function writeAllDocumentsImpl(inputStream,jobOptions) {
 
   let path = '/v1/internal/forestinfo';
   let connectionParams = this.client.getConnectionParams();
   let requestOptions = mlutil.copyProperties(connectionParams);
 
-  let inputStream = new stream.PassThrough({objectMode: true});
   requestOptions.method = 'GET';
   requestOptions.headers = {
     'Accept':       'application/json',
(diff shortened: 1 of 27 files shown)


test-basic/documents-data-movement-readAll.js
```diff
@@ -496,23 +497,23 @@ describe('data movement readAll', function() {
...
         uriStream.push(null);
-        const readAllStream = dbWriter.documents.readAll({categories: ['rawContent','permissions']});
+        const readAllStream = dbWriter.documents.readAll(uriStream,{categories: ['rawContent','permissions']});
         readAllStream.on('error', function(err){
             err.toString().should.equal('Error: Categories should not have other option(s) if rawContent is needed.');
             done();
         });
 
-        uriStream.pipe(readAllStream);
+      //  uriStream.pipe(readAllStream);
```
rjrudin, line 506: If this line isn't needed anymore, it's better to delete it so there's not confusion later on over whether it's needed or not.

test-basic/documents-data-movement-queryAll.js
```diff
@@ -31,6 +31,7 @@ const query = q.where(ctsQb.cts.directoryQuery('/test/dataMovement/requests/quer
 
 describe('data movement queryAll', function() {
     before(function (done) {
+        this.timeout(5000);
```
rjrudin, line 34: Timeouts in tests are anti-patterns, as there's no guarantee a break won't occur. If the thing we want to test is "Make sure we can query all of the URIs that match a query", then testing that we can get more than 1 URI back - i.e. 2 - is typically as effective as testing that we can get 10k URIs back. You may need a few more than 2 URIs - i.e. 10 - for some of the test cases in this module, but I definitely would not do 10k, there's no reason to spend time writing and querying 10k docs when 10 should do just fine. 

While it may not be possible to remove the timeout, writing/querying 10 docs should allow the timeout to be both set fairly low and never exceeded.

lib/documents.js
```diff
@@ -2275,18 +2275,17 @@ Documents.prototype.suggest = function suggestDocuments() {
  * @returns {stream.Writable} - a stream.Writable in object mode that receives document descriptor input from the
  *  application for writing to the database.
  */
-Documents.prototype.writeAll = function writeAllDocuments(options) {
+Documents.prototype.writeAll = function writeAllDocuments(stream,options) {
```
rjrudin, line 2278: These all look like breaking changes - technically fine, since this will be in a major release - since a new argument is being added to the beginning of the list of arguments. What's the blast radius of this though? i.e. in addition to mentioning this in release notes, how many changes will be needed to the docs to ensure they don't show any examples from the previous way of invoking these methods?

test-basic/documents-data-movement-queryAll.js
anu3990: Hi Rob, yes lesser number of docs are working too. I added this since i wanted to test with larger number of docs. This timeout is allows the docs to be written to the database before being read. Mocha has a built-in timeout of 2 seconds and without timeout, the tests started reading the docs before all the docs were written and threw an error.

lib/documents.js
anu3990: The radius is big. I will work on the functional tests and examples next and also with the doc team to update the documentation.

test-basic/documents-data-movement-readAll.js
anu3990: Done.

test-basic/documents-data-movement-readAll.js
```diff
@@ -154,9 +155,9 @@ describe('data movement readAll', function() {
         }
```
anu3990, line 155: @georgeajit this test might still fail. I am working on it.

anu3990: Fixed. It should no longer fail. Thanks.!

PR #722, opened 2023-02-24
NODE-121 : Connection error causes Node process to terminate

anu3990:


mariusmarklogic (approved):

rjrudin (changes requested):
I am assuming you can add a test for this to trigger the error condition - I haven't tried it, but perhaps send a request that uses e.g. certificate auth to the app server that expects digest.

PR #724, opened 2023-03-06
Updating CHANGELOG.md and package.json

anu3990:


georgeajit (changes requested):
Explanation for Git issue 540 needs to match other items in terms upper case/lower case

georgeajit (approved):

PR #730, opened 2023-03-13
Fix for Jenkins failures.

anu3990:


lib/rows.js
```diff
@@ -81,9 +81,9 @@ Rows.prototype.query = function queryRows() {
   }
 
   let builtPlan = args[0],
-      options = (args.length === 2) ? args[1] : null;
+      options = (args.length >= 2) ? args[1] : null;
 
-  return queryRowsImpl(this, builtPlan, null, options, args[1]);
+  return queryRowsImpl(this, builtPlan, null, options, args[2]);
 };
```
anu3990, line 87: @mariusmarklogic , we might need to use args[2] instead of args[1] since args[1] is expected to be sent by plan-search.js and other test failures on jenkins

test-basic/bindingFromParam.js
```diff
@@ -408,7 +409,7 @@ describe('binding from param', function () {
         ];
```
anu3990, line 409: @mariusmarklogic  , this test is still failing. Can you please checkin dataFromParam.json file in the test-basic/data folder.

mariusmarklogic (approved):

PR #740, opened 2023-04-11
DEVEXP-354 : Refresh expired token when using MarkLogic cloud authention

anu3990:


lib/requester.js
```diff
@@ -160,7 +162,7 @@ function startRequest(operation) {
       throw new Error('request for unknown authenticator type '+authType);
     }
   } else {
-    if(operation.client.connectionParams.apiKey && !operation.accessToken){
+    if(operation.client.connectionParams.apiKey && (!operation.accessToken || operation.expiration <= (new Date()))){
```
BillFarber, line 165: What happens if there's a race condition and the accessToken expires in the time between line 165 and when the server receives the request?

rjrudin: Right, I think there are windows where this won't be reliable. 

The approach in the Java Client is to not bother with the expiration time and just assume that the first 403 it receives is due to the token expiring. It then generates a new token and tries the request again. If that also results in a 403, then it propagates the 403 to the user.

Here's the code for checking on a 403 - https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/impl/okhttp/MarkLogicCloudAuthenticationConfigurer.java#L160 .

And then assuming that multiple threads can be using the same client instance (I assume that's possible in the DMSDK impl in the Node Client), you should synchronize the code block that generates the new token as shown here - https://github.com/marklogic/java-client-api/blob/develop/marklogic-client-api/src/main/java/com/marklogic/client/impl/okhttp/MarkLogicCloudAuthenticationConfigurer.java#L183 . That accounts for 2 or more threads determining that a new token is needed at the same time - we only want 1 of those threads to "win" and generate a new token, and then the other threads will realize the token has been updated and will use the new token.

rjrudin (changes requested):
Is there a way use mocks in the tests in the Node Client? I used the mock HTTP web server offered by OkHttp in the Java Client to emulate 403's being returned to the client. Without that, it's going to be very difficult to test this code - even doing it manually right now will be difficult.

lib/requester.js
anu3990: Let me update the errorListener as well

anu3990:
I will test this manually in two weeks when my current access token expires.

lib/responder.js
```diff
@@ -1156,6 +1157,17 @@ function operationResultStream() {
 function operationErrorListener(error) {
   /*jshint validthis:true */
   var operation = this;
+  if(operation.client.connectionParams.apiKey){
+    if(error.statusCode === 403 && operation.expiration <= (new Date())){
```
rjrudin, line 1161: Is the check on `expiration` to avoid an endless loop where a 403 is returned for some reason other than an expired token? Otherwise, I think it's better to just check for a 403 - that's what the Java Client and XCC are doing.

lib/responder.js
```diff
@@ -1156,6 +1157,17 @@ function operationResultStream() {
...
+      if(!operation.lockAccessToken){
```
rjrudin, line 1162: I am now wondering if anything is needed - according to https://stackoverflow.com/a/7238663/3306099 , JavaScript doesn't have the same issues as e.g. Java where multiple threads can access the same block of code at the same time. If that's the case, then `lockAccessToken` isn't needed - the code can just say "If I get a 403, then I'll generate a new token and try again".

lib/responder.js
anu3990: Yes, it is to avoid loop.

lib/responder.js
anu3990: That's correct. This condition will be helpful for DMSDK where we are creating multiple threads.

lib/requester.js
```diff
@@ -71,11 +70,17 @@ function getAccessToken(operation, callback){
...
+      operation.accessToken = responseValue.access_token;
+      operation.expiration = new Date(responseValue[".expires"]);
+      if(operation.lockAccessToken){
+        operation.lockAccessToken = false;
+      }
+      authenticatedRequest(operation);
     });
     res.on('error', (e) => {
       operation.accessToken = new Error(e);
+      throw new Error(operation.accessToken);
```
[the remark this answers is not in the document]
rjrudin: I think that's worth a comment to explain what's going on there - e.g.

    // Throw an error that the MarkLogic Cloud authentication code will recognize

I'm not clear though on how that works - line 304 is doing this:

    if (operation.accessToken instanceof Error)
        throw new Error(operation.accessToken);

If an error is now thrown, then under what condition would `operation.accessToken` be an instance of Error?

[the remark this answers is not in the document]
anu3990: On line 82, operation.accessToken is set as the Error. So for a thread that came here (since it did not get the getAccessToken lock or for whatever reason), will be notified not to proceed with the operation.

rjrudin (approved):
Seems fine, and like the Java Client and XCC, we won't really know if it works until we have a more reliable way of testing expired tokens.

PR #741, opened 2023-04-13
DEVEXP-289: The "remove" optic function throws a multipart error

mariusmarklogic:
the tests are passing as expected in windows machine
image

### DIFF
--- a/lib/responder.js
+++ b/lib/responder.js
@@ -471,6 +471,30 @@ MultipartDispatcher.prototype.promise = function dispatchMultipartPromise(
   var operation = this.operation;
   operation.logger.debug('multipart promise');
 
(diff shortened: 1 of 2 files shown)


anu3990 (changes requested):
@mariusmarklogic , i ran the below test and with the fix in this PR it does not theow error but it also does not return the uris of the documents deleted. -
it('with query remove', function(done){
const op = marklogic.planBuilder;
const planFromBuilderTemplate = op.fromDocUris(op.cts.directoryQuery('/test/optic-update/')).remove();
db.rows.query(planFromBuilderTemplate)
.then(function(response)

{ console.log(response); done(); }
)
.catch(err=> done(err));
});

anu3990 (approved):

PR #742, opened 2023-04-19
DEVEXP-362 : Submit GraphQL query using Node Client

anu3990:


BillFarber (approved):

mariusmarklogic (dismissed):

lib/rows.js
```diff
@@ -160,9 +160,9 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
     }
   }
 
-  const contentTypeHeader = queryContentType(builtPlan, options.queryType);
+  const contentTypeHeader = graphqlQuery?'application/graphql':queryContentType(builtPlan, options.queryType);
```
rjrudin, line 163: I think both this and the line below it would benefit from spaces for readability - e.g. 

    const contentTypeHeader = graphqlQuery ? '...' : queryContentType(...);

I also often like doing this for ternary operations:

    const contentTypeHeader = graphqlQuery ? 
        "application/graphql" : 
        queryContentType(builtPlan, options.queryType);

I usually put it on one line if the line is short enough, and then 3 lines if one of the two results is a little lengthy.

lib/rows.js
```diff
@@ -264,8 +264,11 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
   );
   operation.validStatusCodes  = [200, 404];
   operation.complexValues  = complexValues;
+  if(graphqlQuery){
```
rjrudin, line 267: I think spaces are needed here too - it looks like most of the code in this module has standard spaces before and after the parents. Is there a linter for node-client-api that would catch things like this?

rjrudin (changes requested):

anu3990:
Fixed the spaces.

rjrudin (approved):

PR #745, opened 2023-05-02
DEVEXP-287: lockForUpdate implementation in node-client-api

mariusmarklogic:


test-basic/lockForUpdate-test.js
```diff
@@ -0,0 +1,110 @@
...
+                (duration > 1500).should.equal(true);
+                db.documents.read({uris: [uri], categories: ['metadata']})
+                    .result(function (documents) {
+                        try {
+                            documents.length.should.equal(1);
+                            documents[0].collections.should.deepEqual(['optic']);
+                            done();
+                        } catch (e) {
+                            done(e);
+                        }
```
anu3990, line 61: Please add the above in "then" block so that it is not read before it is updated.

anu3990 (dismissed):

PR #750, opened 2023-05-09
DEVEXP-444: Setting "rejectUnauthorized" doesn't seem to work with a self-signed certificate

anu3990:


BillFarber (approved):

lib/marklogic.js
```diff
@@ -738,7 +738,7 @@ function initClient(client, inputParams) {
   for (var i=0; i < keys.length; i++) {
     var key = keys[i];
     var value = inputParams[key];
-    if (value) {
+    if (value != null) {
```
rjrudin, line 741: This should at a minimum have a test that would fail without the change and now passes with the change. 

The `initClient` function is fairly long - about 80 lines of code - and is thus difficult to test and understand. So extracting a function here would have two benefits - 1) initClient would be easier to understand, and 2) we can more easily test this change. 

It looks like the first 30 lines are focused on building `connectionParams`. So an obvious function would be:

    function buildConnectionParams(inputParams);

And that would return the `connectionParams` object. Unit tests can then easily be written to ensure that the function is behaving correctly - including the fix above.

BillFarber:
I expect this to work, but can we add a test to verify. That will also help prevent changing it back again.

rjrudin (changes requested):

lib/marklogic.js
rjrudin: It may be better to have `buildConnectionParams` instead be `initConnectionParams` and only contain the first 22 lines - everything through the `for` loop that is modified here.

rjrudin (approved):

BillFarber (approved):

PR #765, opened 2023-05-24
DEVEXP-415 : Call any REST endpoint with Node Client

anu3990:


lib/internal.js
```diff
@@ -0,0 +1,52 @@
...
+
+    function sendRequest(userOptions, optionalName,  requestType='single', responseType='single'){
+        if(!userOptions.path){
+            throw new Error('Path is needed to send request.');
+        }
+        mlutil.copyProperties(userOptions,requestOptions);
+        let operation = new Operation(
+            optionalName?optionalName:'', clientObject, requestOptions, requestType, responseType
+        );
+        operation.validStatusCodes = [200, 201, 204, 404];
```
rjrudin, line 38: This looks like it needs to be another arg to `sendRequest` with a default value of `[200, 201, 204, 404]`. Those may not be the correct codes for any given endpoint, so the user needs to be able to override them. 

So perhaps:

    sendRequest("/v1/ping", myOptions, null, "single", "single", [200, 204]);

test-basic/test-internal.js
```diff
@@ -0,0 +1,80 @@
...
+
+    it('test newRequestOptions creation', function(done){
+        const requestOptions = dbWriter.internal.newRequestOptions('/v1/ping');
+        requestOptions.method = 'POST';
+        requestOptions.headers = {
+            'Accept':       'application/json',
+            'Content-Type': 'application/json'
+        };
+        try {
+            should.deepEqual(requestOptions.path, '/v1/ping');
```
rjrudin, line 48: I think the main thing to test here is that the optional "database" and "basePath" params are accounted for. I don't think you need to test the values of "method" and "headers", as the user is setting those themselves. But the user would expect "path" to account for "database" and "basePath" so that the user doesn't need to account for them.

lib/internal.js
anu3990: But what if the user inputs 400 or 500 as validStatusCodes?

rjrudin: That's a user decision and it's up to the user to pick sensible status codes. The more likely use case is the user only wants to specify a single status code as a valid response, such as `[200]`.

test-basic/test-internal.js
anu3990: Adding a new test for this.

rjrudin (commented):
LGTM, but I defer to @BillFarber on whether it will handle all the scenarios it's needed for in mlxprs.

BillFarber:
> 

LGTM - I don't think I'm smart enough to be sure it will handle all the required scenarios. I think this is a good place to start and then if we need changes we'll create new stories.

BillFarber (approved):

PR #768, opened 2023-06-01
Maybe this would work

rjrudin:


lib/internal.js
```diff
@@ -27,15 +27,20 @@ let clientObject;
         return this;
     }
 
-    function sendRequest(userOptions, optionalName,  requestType='single', responseType='single', validStatusCodes = [200, 201, 204, 404]){
+    function sendRequest(userOptions, operationCallback, optionalName, requestType='single', responseType='single'){
```
rjrudin, line 30: I am thinking the callback should be the second arg, assuming it's likely to populate it. That allows for omitting the last 3 args (if "single" works for both the request and response type).

lib/internal.js
```diff
@@ -27,15 +27,20 @@ let clientObject;
...
         }
         mlutil.copyProperties(userOptions,requestOptions);
         let operation = new Operation(
             optionalName?optionalName:'', clientObject, requestOptions, requestType, responseType
         );
-        operation.validStatusCodes = validStatusCodes;
+        if (operationCallback) {
+            operationCallback(operation);
+        }
+        if (!operation.validStatusCodes) {
```
rjrudin, line 41: The thought is that if the user doesn't provide validStatusCodes, we'll default to these.

PR #771, opened 2023-06-05
Updating version and CHANGELOG.md

anu3990:


CHANGELOG.md
```diff
@@ -1,11 +1,25 @@
 # CHANGELOG
 
+## 3.1.0
+#### New Functionality
+
+- Support for Optic-update Apis.
+- New endpoint for GraphQL.
+- Support for calling any REST endpoint with Node Client.
+- Support to refresh expired token when using MarkLogic cloud authentication.
```
rjrudin, line 9: I'm being picky, but I like to avoid "Supports for" or "Support to" in favor of a more active description - i.e. "Tokens are automatically refreshed before expiration when using MarkLogic Cloud authentication."

That makes it clear too that the client is doing the work, whereas "support for" implies that a user may need to do something to leverage the new "support".

CHANGELOG.md
```diff
@@ -1,11 +1,25 @@
...
+- Fix for security vulnerabilities.
+
+
 ## 3.0.0
 #### New Functionality
 
 - [#614](https://github.com/marklogic/node-client-api/issues/614) - Exporting Rows - queryAll api.
 - [#644](https://github.com/marklogic/node-client-api/issues/644) - Reprocessing documents - transformAll and queryToTransformAll apis.
 - [#665](https://github.com/marklogic/node-client-api/issues/665) - Deleting Documents  - removeAllUris and queryToRemoveAll api.
+- Support to connect to MarkLogic cloud using ApiKey.
```
rjrudin, line 22: Should be "apiKey" instead of "ApiKey". And capitalize "Cloud" as well.

rjrudin (changes requested):

rjrudin (approved):

PR #776, opened 2023-06-14
DEVEXP-483 : Filter Node Client tests if they are not supposed to run on server version less than 11

anu3990:


anu3990:
@mariusmarklogic , the changes here are mostly on optic-update tests. This is a jenkins run for this PR and the test failures look better - http://jenkins1.marklogic.com:9000/blue/organizations/jenkins/Node-client-api/detail/PR-773/5/tests/
Please let me know if I missed out any file. Thanks.!

test-basic/bindingFromParam.js
```diff
@@ -20,9 +20,33 @@ const db = marklogic.createDatabaseClient(testconfig.restWriterConnection);
...
 const fs = require("fs");
+const testlib = require("../etc/test-lib");
+let serverVersionGreaterThanEqual11 = false;
 
-describe('binding from param', function () {
-    this.timeout(20000);
+describe('optic-update fromParam tests', function(){
+
+    before(function (done) {
+        const manageClient =
```
rjrudin, line 29: This is a lot of duplication of the same code - couldn't this all become a function in test-lib, like `requiresMarkLogic11()`?

anu3990: Added findServerConfiguration in test-lib. We will also be able to use it later for tests related to features going in MarkLogic 12

mariusmarklogic (approved):

test-basic/bindingFromParam.js
```diff
@@ -20,9 +20,26 @@ const db = marklogic.createDatabaseClient(testconfig.restWriterConnection);
...
+const testlib = require("../etc/test-lib");
+let serverVersionGreaterThanEqual11 = false;
 
-describe('binding from param', function () {
-    this.timeout(20000);
+describe('optic-update fromParam tests', function(){
+
+    before(function (done) {
+        testlib.findServerConfiguration().result(function (response) {
+            serverVersionGreaterThanEqual11 = (parseInt(response.data['local-cluster-default'].version) >= 11);
```
rjrudin, line 30: I think this is still worth putting into a helper function; it avoid duplicating both code and knowledge of the "local-cluster-default/version" path. And it would make the code more expressive - e.g. "isMarkLogic11OrHigher()". It's also likely that there will be future methods like this - e.g. "isMarkLogic12OrHigher()".

anu3990: I think it should be okay. "findServerConfiguration" is a generic function. Whether the tests need to run on MarkLogic 11 or 12 is purely a test thing. For example, in the future we add a new test file - future_tests.js and it has few tests that should run on 11 or higher and few tests that should run on 12 or higher. In that case, a single call to this generic function would suffice and then the test file can filter out. We won't be needing two different function calls there.
For the "local-cluster-default/version", i did try to filter out the version in "findServerConfiguration" but there were issues caused by "done" being called before resolving the promise (if i remember correctly)

PR #785, opened 2023-07-06
DEVEXP-276 : Refactor "databaseParam" in mlutil.js

anu3990:


BillFarber (approved):
All the changes look straight-forward.
Approving with the assumption that the failures are not related to these changes. Please verify that fact.

lib/documents.js
```diff
@@ -269,7 +269,7 @@ function probeDocumentsImpl(contentOnly, args) {
   var connectionParams = this.client.getConnectionParams();
   var requestOptions = mlutil.copyProperties(connectionParams);
   requestOptions.method = 'HEAD';
-  requestOptions.path = mlutil.databaseParam(connectionParams, path, '&');
+  requestOptions.path = mlutil.newRequestOptions(connectionParams, path, '&');
```
rjrudin, line 272: This seems unexpected - the function is called `newRequestOptions`, implying that it will return a "request options" object, but its return value is instead being assigned to the value of "path". 

I think what's shown in the JIRA ticket reads better and also removes more code:

```
const requestOptions = mlutil.newRequestOptions(
    connectionParams, path, '&');
```

That would replace line 270 above as well, and it would return a requestOptions object that has the path correctly constructed (which includes accounting for the optional "database" param). Each client of that function would then assign `method` and any other properties necessary.

rjrudin (changes requested):

lib/documents.js
rjrudin: Note too that since `method` seems to always be assigned in this repeated block of code, it would be reasonable to add that as a param:

```
const requestOptions = mlutil.newRequestOptions(connectionParams, path, 'POST', '&');
```
```

Comment (anu3990) [reply lib/documents.js:272]: you mean we should include the copyProperties ?

Comment (anu3990) [reply lib/documents.js]: Okay so changing databaseParam -> newRequestOptions(connectionParams, endpoint, separator, method) that sets the method as well.

File: lib/documents.js [hunk 2]
```diff
@@ -267,9 +267,7 @@ function probeDocumentsImpl(contentOnly, args) {
   }
 
   var connectionParams = this.client.getConnectionParams();
-  var requestOptions = mlutil.copyProperties(connectionParams);
-  requestOptions.method = 'HEAD';
-  requestOptions.path = mlutil.databaseParam(connectionParams, path, '&');
+  let requestOptions = mlutil.newRequestOptions(connectionParams, path, '&', 'HEAD');
```
rjrudin, line 270: Sorry, two more small tweaks:

1. It looks like in all these cases, `connectionParams` is only used once. Because fewer variables is almost always better (one exception is when introducing a variable name improves readability, but that's not the case here since it's already clear what `getConnectionParams()` refers to), I would just call `this.client.getConnectionParams()` on line 270 here, and in all the other instances of this code block too. 
2. Unless `requestOptions` is going to be re-assigned to some other value, use `const`. That improves readability by making it clear that `requestOptions` in fact won't be re-assigned to some other value.

rjrudin (changes requested):

lib/documents.js
rjrudin: Right, that's the goal of the refactor - try to condense a block of 4 or 5 lines that is frequently repeated into a single readable line of code.

lib/documents.js
anu3990: Done

lib/internal.js
```diff
@@ -31,7 +31,7 @@ let clientObject;
         if(!path){
             throw new Error('Path is needed to send request.');
         }
-        requestOptions.path = mlutil.databaseParam(clientObject.getConnectionParams(), path, '?')
+        requestOptions.path = mlutil.newRequestOptions(clientObject.getConnectionParams(), path, '?').path;
```
rjrudin, line 34: This looks like it can be simplified by changing the `initialize` method (which is misspelled, it's missing an "i") to not instantiate the `requestOptions` object. So remove line 26 and then change this to :

    requestOptions = mlutil.newRequestOptions(connection params, path , '?'); 

That also allows for line 21 to be removed - I don't think there's any benefit to keeping `requestOptions` around as state, it's simpler to create it each time that `sendRequest` is invoked.

lib/rows.js
```diff
@@ -244,6 +242,8 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
       endpoint += (complexValues === 'reference') ? sep + 'node-columns=reference' : '';
       break;
   }
+  // Specify database if defined in client connection
```
rjrudin, line 245: This comment looks obsolete now and should be removed.

lib/documents.js
```diff
@@ -2702,7 +2654,7 @@ function collectDocUris(forestName, jobState){
     operation.timestamp = new mlutil.Timestamp(null);
     operation.jobState = jobState;
   }
-  requestOptions.path = mlutil.databaseParam(connectionParams, forestEndpoint);
+  requestOptions.path = mlutil.newRequestOptions(connectionParams, forestEndpoint).path;
```
rjrudin, line 2657: This is pretty tricky to make sense of - it raises the question of "Why are we calling `newRequestOptions` but only using the `path`?" 

Doesn't need to be changed now, but it looks like this `collectDocUris` method can be simplified and made more readable by doing the following:

1. Determine `forestEndpoint` as early as possible. It looks like it can be determined based on the `jobState`. 
2. Call `newRequestOptions` as late as possible (right before the `Operation` is instantiated), passing in `forestEndpoint` as the `path` param. 

That should result in a simpler flow:

1. The `forestEndpoint` is calculated.
2. `newRequestOptions` is invoked. 
3. The `Operation` is instantiated. 
4. The `operation` instance is modified as needed. 
5. `startRequest` is invoked. 

You could try doing that here, but I think it's fine to tackle in a separate refactor later on.

rjrudin (changes requested):
Looks good, just one had idea for simplifying things in internal.js - the comment in documents.js can be addressed later on .

lib/internal.js
anu3990: This ticket was for refactoring "databaseParam" not internal.js. This file has few problems while refactoring but I did not want to change lot of Phil's code. We can open a new Jira ticket.

lib/rows.js
anu3990: I can remove this and also rename "initialize" in internal.js but giving review comments together will save time in the future for both reviewer and developer.

lib/documents.js
anu3990: I tried doing it different ways but since "forestEndpoint" changes based on conditions, this was the only way it was working.

lib/internal.js
rjrudin: We'll use a separate ticket to address `let requestOptions`.

rjrudin (approved):

PR #787, opened 2023-07-11
DEVEXP-502 : In the Node client: internal.js has two global variables that should be refactored

anu3990:
…

### DIFF
--- a/test-basic/test-internal.js
+++ b/test-basic/test-internal.js
@@ -18,6 +18,7 @@ const testconfig = require("../etc/test-config");
 const marklogic = require("../lib/marklogic");
 const should = require('should');
 let dbWriter = marklogic.createDatabaseClient(testconfig.restWriterConnection);
+let assert = require('assert');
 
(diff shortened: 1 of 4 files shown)


test-basic/test-internal.js
```diff
@@ -94,17 +95,42 @@ describe('internal tests', function() {
...
-                should.exist(response.start);
-                should.exist(response['page-length']);
-                for(let i=0; i<response.length; i++){
-                    should.exist(response.results[i]);
-                }
+                let stringResponse = response.toString();
+                assert(stringResponse.includes("total"));
+                assert(stringResponse.includes("start"));
+                assert(stringResponse.includes("page-length"));
             })
```
anu3990, line 102: Modified this test since we get an xml response.

rjrudin (approved):

BillFarber (approved):
After this is merged into the develop branch, I will test it with MLXPRS.

PR #792, opened 2023-10-04
DEVEXP-587 : Update README to include an example for connecting to ML Cloud

anu3990:


README.md
```diff
@@ -48,6 +48,14 @@ var db = marklogic.createDatabaseClient({
   authType: 'DIGEST'
 });
 
+// For MarkLogic Cloud
+let db = marklogic.createDatabaseClient({
```
rjrudin, line 52: Small thing, but both this and the above example should do `const db` as there will rarely ever be a reason to reassign the variable to something else, and thus `const` is a good practice to encourage. 

Also, I think a more realistic value for "basePath" would be useful here - something like "/ml/manage/test", or you can copy what's at https://github.com/marklogic/ml-gradle/wiki/MarkLogic-Cloud-support .

rjrudin (changes requested):

rjrudin (approved):

PR #793, opened 2023-10-04
DEVEXP-588 : High vulnerability detected in Node Client Api

anu3990:


package.json
```diff
@@ -3,7 +3,7 @@
   "name": "marklogic",
   "description": "The official MarkLogic Node.js client API.",
   "homepage": "https://github.com/marklogic/node-client-api",
-  "version": "3.1.0",
+  "version": "3.2.0",
```
BillFarber, line 6: Should we make it 3.1.1 since there is no new functionality?

package.json
```diff
@@ -78,7 +78,9 @@
     "nanoid": "3.1.31",
     "simple-get": "4.0.1",
     "minimatch": "5.1.0",
-    "semver": "7.5.3"
+    "semver": "7.5.3",
+    "get-func-name": "2.0.1",
```
rjrudin, line 82: Since these weren't listed before, they must have been transitive dependencies. Which package(s) were they being inherited from, and can we update those dependencies instead? 

Otherwise, future developers may wonder why these transitive dependencies are listed here. So at a minimum, the commit message should explain why they have been added - e.g. "get-func-name is normally inherited by package abc, but to address CVE 123, we are now explicitly stating a dependency on it so we can use the patched version".

package.json
rjrudin: If we release from develop, then 3.2.0 is fine, but if we weren't ready to release from develop, then yes, we should do 3.1.1 and toss this PR onto a branch off of master.

rjrudin (changes requested):

rjrudin (approved):

PR #799, opened 2023-10-10
DEVEXP-593 : Filter tests based on server versions.

anu3990:


test-basic/documents-data-movement-rows-queryAll.js
```diff
@@ -35,10 +36,17 @@ const tdeWriter = marklogic.createDatabaseClient({
...
 let result = [];
 const planFromBuilderTemplate = p.fromView('soccer', 'matches', '');
+let serverVersionGreaterThanEqual12 = false;
 
-describe.skip('data movement rows-queryAll', function() {
+describe('data movement rows-queryAll', function() {
     this.timeout(15000);
     before(function(done){
+        testlib.findServerConfiguration().result(function (response) {
+            serverVersionGreaterThanEqual12 = (parseInt(response.data['local-cluster-default'].version) >= 12);
```
rjrudin, line 45: I think given that this check on the version compared to 12 is in several places, and that's true of the check on version 11, test-lib.js should make this easy for all the tests with methods of `testlib.isVersion12OrHigher()` and `testlib.isVersion11OrHigher()`. test-lib.js can potentially cache the results as well so it only needs to get the server version once during an entire test suite run. 

That will avoid duplicating this code here - i.e. getting `local-cluster-default` and parsing it - and will make the tests shorter and more readable.

rjrudin (changes requested):
Looks good, I just think the duplicate checks on server version should be consolidated into test-lib.js.

test-basic/documents-data-movement-rows-queryAll.js
anu3990: This test here checks for server version 12. The other tests check for server version 11. We have added a function to find the server version. The tests can take a decision based on the version returned. If we start adding functions based on server versions we will have a lot of functions. For example, in the future if few features are added to 12.2 and not available in 12.1, we will have to add further functions. This is a better approach where the server version is returned and then the tests can take decision based on that.

rjrudin: Sure, though functions are cheap and self-documenting and avoid duplication. Can you instead add a `testlib.getServerVersion()` function so that we're not repeating the code that parses the version out of the response?  That's really my main concern - the duplication of `parseInt(response.data['local-cluster-default'].version)`.

anu3990: Updated the files and test-lib

etc/test-lib.js
```diff
@@ -195,10 +195,12 @@ function createManager(adminClient) {
...
 
-function findServerConfiguration(){
+function findServerConfiguration(serverConfiguration){
   const manageClient = marklogic.createDatabaseClient(testconfig.manageAdminConnection);
-  return new Manager(manageClient).get({
+    new Manager(manageClient).get({
     endpoint: '/manage/v2?format=json',
+  }).result(function(response){
+      serverConfiguration.serverVersion =  parseFloat(response.data['local-cluster-default'].version).toFixed(2);
   });
```
anu3990, line 204: Currently "serverConfiguration" only has version but as we move ahead, we can add more from the response as needed.

rjrudin (approved):

PR #814, opened 2023-11-29
DEVEXP-596 : Configure gZIP option

anu3990:


test-basic/documents-core.js
```diff
@@ -57,6 +57,40 @@ describe('document content', function(){
...
+                    documents.length.should.equal(1);
+                    let document = documents[0];
+                    document.should.have.property('content');
+                    try{
+                        document.content.should.have.property('key1');
+                    } catch(error){
+                        // An AssertionError should be thrown here since the contents are zipped and cannot have property key1.
+                        done();
+                    }
+                })
```
anu3990, line 75: Verified in responder by logging - response.headers -
{
  'vnd.marklogic.document-format': 'json',
  'content-type': 'application/json; charset=utf-8',
  'ml-effective-timestamp': '17013030096192060',
  server: 'MarkLogic',
  'content-encoding': 'gzip',
  connection: 'close'
}

rjrudin (changes requested):
I don't think this is working quite yet, as the tests I tried resulted in a gzipped response, and I don't think we want to force the user to deal with that themselves - i.e. they expect the compression to be transparent. 

This thread - https://stackoverflow.com/questions/8880741/easy-http-requests-with-gzip-deflate-compression - seems to provide good examples of how to use `zlib` to decompress the response data. Ideally, there's a single spot in the Node Client where a check can be made to see if the "Content-Encoding" header is "gzip", and if so, use zlib to decompress the response data. 

FWIW, this is the test I was using - I was running it against the java-unittest app server that's created by the Java Client test app, but any program like this should work:

```
var marklogic = require('../lib/marklogic.js');

const client = marklogic.createDatabaseClient({
    host: "localhost",
    port: 8012,
    user: "rest-admin",
    password: "x",
    enableGzippedResponses: true
});

client.rows.query(
    "op.fromView('opticUnitTest', 'musician').limit(1)", 
    {queryType: "dsl"}
).then(function(response) {
    console.log("RESULT", response);
});
```

rjrudin:
This may be more helpful - https://stackoverflow.com/a/12776856/3306099 - the previous link I posted seemed to be more focused on the `requests` library and the Node Client is using the `http` library instead.

rjrudin (approved):
Looks good, though I recommend trying to manually run all the tests with gzip enabled too to make sure you get the same results as running them with gzip disabled.

rjrudin:
I tried running all the tests locally. I get 8 failures; I don't know if those are intermittent or not. The failures are:

  1) data movement queryAll
       "before all" hook for "queryAll should throw error if queryType is not cts":
  2) data movement queryAll
       "after all" hook for "queryAll should throw error with consistentSnapshot as Integer":
  3) data movement readAll
       "before all" hook for "should readAll documents with empty options":
  4) data movement rows-queryAll
       should queryAll documents with onCompletion and outputStreamType options:
  5) temporal protect
       should protect a temporal document with a duration and archivePath:
  6) Endpoint caller
       "before all" hook for "moduleInitSjs endpoint":
  7) lexicons
       with nullable column:
  8) Service caller
       "before all" hook for "moduleInitSjs endpoint":
       
But when I change the code to always send the gzip header, I get 164 failures. I'm going to try some of those individual tests with gzip enabled as opposed to always enabling it.

rjrudin:
I was able to narrow it down a bit - if I run just optic-update-execute.js (I could only do this by modifying the "test" function in gulpfile.js), 3 of the 4 tests fail - e.g.

```
1) optic-update execute tests
       execute 
         test with fromDocUris -> remove:
     Uncaught SyntaxError: Unexpected token  in JSON at position 0
      at JSON.parse (<anonymous>)
      at /Users/rudin/workspace/node-client-api/lib/responder.js:504:29
```

So the code in responder.js may need an enhancement to handle empty responses or something like that.

rjrudin (changes requested):
I think I found the problem. I tried running only optic-update-execute.js, and only one of the tests in there (I temporarily removed the other 3 to make it fast, so I was only running "test with fromDocUris -> remove"). 

I think line 504 needs a modification:

    let data = JSON.parse(chunks);

It is receiving the compressed gzip here and thus throwing an error. I don't know what the purpose of this code is - it's only used for Optic queries based on the "errorListenerCheck" condition. Presumably, it needs to run after the code you added for decompressing the gzipped response.

lib/responder.js
```diff
@@ -378,11 +379,25 @@ BodyDispatcher.prototype.promise = function dispatchBodyPromise(
 
   var isString = operation.copyResponseHeaders(response);
 
-  // concatStream accumulates response with callback
-  response.pipe(concatStream(
-      {encoding: (isString ? 'string' : 'buffer')},
-      collectObject
-      ));
+  if(response.headers['content-encoding']==='gzip') {
```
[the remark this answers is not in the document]
rjrudin: I think a function is in order as this comparison appears 13 times now. `if (responseIsGzipped(response))` would read nicely and would eliminate the duplication. It could also easily check for "content-encoding" and "Content-Encoding" as well.

test-basic/documents-negative.js
```diff
@@ -112,12 +112,16 @@ describe('document negative', function(){
...
-      done();
+      try {
+        error.should.have.property('body');
+        error.body.should.have.property('errorResponse');
+        error.body.errorResponse.should.have.property('message');
+        error.body.errorResponse.message.should.containEql('XDMP-RWINVAL');
+        error.body.errorResponse.message.should.containEql('set-transaction');
+        done();
+      } catch(error){
+        done(error);
```
rjrudin, line 123: Does this result in the test failing - i.e. the error is caught, and then `done(error)` results in a failed test with the message from `error`?

rjrudin (commented):
Looks good, just had a comment about removing some duplication and wondering why the try/catch blocks are needed for in the tests.

test-basic/documents-negative.js
anu3990: This will catch any assertions errors thrown from line 116-120. Without the try-catch block, if there were any assertion failures, the done would not be called and the test-run would show timeout. This way, we can see the actual error.

rjrudin (approved):

PR #818, opened 2023-12-13
DEVEXP-592 : Submit Optic plan containing new operations in MarkLogic 11.1

anu3990:


lib/plan-builder.js
```diff
@@ -84,6 +84,9 @@ class Builder extends bldrgen.PlanBuilder {
     if(this.rdt === void 0) {
       this.rdt = new RdtExpr();
     }
+    if(this.ctsString === void 0){
+      this.ctsString = new CtsExpr();
```
rjrudin, line 88: I think we need to take the same approach in the Java Client and modified the generated `CtsExpr` class in plan-builder-generated.js. While that's not desirable, I think it's better to provide a simple, consistent UX for a developer rather than to complicate the UX by introducing `ctsString` - we're making things more complicated for the user solely because we have a buggy code generator. Until the code generator is fixed, the right solution is to modify the buggy generated code.

rjrudin (changes requested):

lib/plan-builder-generated.js
```diff
@@ -1115,7 +1115,7 @@ point(...args) {
     const paramdefs = [['latitude', [types.XsDouble, PlanColumn, PlanParam], true, false], ['longitude', [types.XsDouble, PlanColumn, PlanParam], true, false]];
     const checkedArgs = (namer !== null) ?
         bldrbase.makeNamedArgs(namer, 'cts.point', 2, new Set(['latitude', 'longitude']), paramdefs, args) :
-        bldrbase.makePositionalArgs('cts.point', 2, false, paramdefs, args);
+        bldrbase.makePositionalArgs('cts.point', 1, false, paramdefs, args);
```
rjrudin, line 1118: Sorry but one last thing - any manual modifications you make to this file should have a comment on them explaining that they were manually made. That way, when we run the (hopefully not buggy) generator next time, we'll be able to see in the diff that our manual code is being removed, but the comment on it will tell us not to remove that code.

rjrudin (approved):
Looks good, just have one comment to address about adding comments on manual modifications to the generated code.

rjrudin (approved):

PR #824, opened 2024-01-03
This works locally...

BillFarber:


rjrudin (approved):
LGTM, double-check with @SameeraPriyathamTadikonda  that the JAVA_HOME is suitable.

PR #828, opened 2024-01-05
MLE-4118 : Submit Optic Update DSL plan via Node Client

anu3990:


test-basic/lockForUpdate-test.js
```diff
@@ -98,8 +98,8 @@ describe('optic-update lockForUpdate tests', function() {
         it('test with fromParam with custom uri', function (done) {
             const rows = [{myUri: '/optic/test/musician1.json'}];
             const outputCols = [{"column": "myUri", "type": "string", "nullable": false}];
-
-            db.rows.query(op.fromParam('bindingParam', null, outputCols).lockForUpdate(op.col('myUri')), null, {bindingParam: rows}).then((res) => {
+            const options = serverConfiguration.serverVersion <= 11.1 ? null:{ opticResponseStructure: 'object'};
```
rjrudin, line 101: I don't think you need to introduce `opticResponseStructure` - the Node Client already supports `structure` as an option, you should be able to use that. 

For example, this test in rows.js is doing just that:

```
it('as a JSON array', function(){
  return db.rows.query(planFromJSON, {queryType: 'json', format: 'json', structure: 'array'})
    .then(function(response) {
      valcheck.isArray(response).should.equal(true);
    });
});
```

lib/rows.js
```diff
@@ -274,6 +272,17 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
 return operation;
 }
 
+function createRowsEndpoint(builtPlan, options, graphqlQuery) {
+  let updateFunctions = new Set().add('write').add('lock-for-update').add('remove');
```
rjrudin, line 276: We can't rely on this because it will only work for serialized plans. You should add a test that submits an Optic DSL statement that has write/remove/lockForUpdate in it - this logic won't work for it. 

Instead, per the design in the JIRA ticket, we need the user to tell us that they're doing an update, as we can't reliably determine that otherwise. That's the approach we're taking in the Java and Python clients too. The user can do that via an option - e.g:

    db.rows.query(planOrDslStatement, {"update": true})

To get data back, the user can do:

    db.rows.query(planOrDslStatement, {"update": true, "structure": "object"})

rjrudin (changes requested):

lib/rows.js
```diff
@@ -274,6 +272,15 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
 return operation;
 }
 
+function createRowsEndpoint(options, graphqlQuery) {
+  let endpoint = graphqlQuery?'/v1/rows/graphql':'/v1/rows';
+  if(options && options.update){
+    endpoint = options.structure?'/v1/rows/update?output='+options.structure:
```
rjrudin, line 278: I think `structure` should be getting applied elsewhere already, since it's an existing feature. Try removing that from here, it shouldn't be needed. It should suffice to just determine which of the 3 endpoints are needed here.

rjrudin (changes requested):
Looks good but trying remove the addition of `structure` - I'm thinking that has to be getting applied somewhere already.

lib/rows.js
anu3990: Yes structure is used elsewhere as well. Here, it is used to determine the return type of the response (output=array/object). If we remove it we will not get any response. A better approach would be to use a different variable so that in future if "structure" has other values apart from array and object, optic-update would not be affected.

anu3990: Merging this in. Created a new ticket for making /v1/rows/update?output=object as default unless the user overrides it or if execute is used. Making '/v1/rows/update?output=object' results in failure of few remove and transformDoc tests that might need further investigation.

anu3990:
Merging this in. Created a new ticket for making /v1/rows/update?output=object as default unless the user overrides it or if execute is used. Making '/v1/rows/update?output=object' results in failure of few remove and transformDoc tests that might need further investigation.

PR #829, opened 2024-01-09
MLE-11546 : Default optic update path should be /v1/rows/update?output=object

anu3990:


lib/rows.js
```diff
@@ -272,11 +272,10 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
 return operation;
 }
 
-function createRowsEndpoint(options, graphqlQuery) {
+function createRowsEndpoint(options, graphqlQuery, structure) {
   let endpoint = graphqlQuery?'/v1/rows/graphql':'/v1/rows';
   if(options && options.update){
-    endpoint = options.structure?'/v1/rows/update?output='+options.structure:
-        '/v1/rows/update';
+    endpoint = (options.opticExecute)?'/v1/rows/update':'/v1/rows/update?output='+structure;
```
rjrudin, line 278: IIUC, I think this needs a comment to explain that a user is not expected to set `opticExecute` in the `options` object, but rather it is like an "internal flag" that is used by `rows.execute` to ensure that `output` doesn't default to anything. Otherwise, this looks like it could be part of the public API and something that needs to be documented.

rjrudin (approved):
LG, just recommending that a comment be added.

rjrudin (approved):

PR #830, opened 2024-01-10
Fix for test failures when output=object was added twice.

anu3990:


lib/rows.js
```diff
@@ -209,12 +209,24 @@ function queryRowsOperationImpl(self, builtPlan, streamType, options, bindingArg
     case 'xml':
       acceptHeader = 'application/xml';
       responseType = 'single';
-      endpoint += sep + 'output=' + structure + '&column-types=' + columnTypes;
+      if(options && options.update){
```
rjrudin, line 212: This looks good, but to avoid the increase in duplication code, make a function to handle this - e.g. 

```
function addOutputAndColumnTypes(endpoint, options, structure, columnTypes)
```

rjrudin (changes requested):

PR #838, opened 2024-02-02
MLE-12173 Fixed a bug where a database property caused a problem when creating a URL path for querying ML.

BillFarber:


lib/graphs.js
```diff
@@ -373,7 +373,7 @@ function changeGraph(action, isStreaming, args) {
     if (sep === '?') { sep = '&'; }
   }
 
-  const requestOptions = mlutil.newRequestOptions(this.client.getConnectionParams(), endpoint, sep, ((action === 'write') ? 'PUT' : 'POST'));
+  const requestOptions = mlutil.newRequestOptions(this.client.getConnectionParams(), endpoint, ((action === 'write') ? 'PUT' : 'POST'));
```
rjrudin, line 376: Looks like the `sep` code above this and other places in this file can be removed too?

rjrudin (approved):
All looks good, just wondering if there's a little more code that can be deleted.

lib/graphs.js
BillFarber: Decided against.

PR #843, opened 2024-02-05
Fix for test failure in 11.1.0 vs 11.2-nightly

anu3990:


rjrudin (approved):

test-basic/lockForUpdate-test.js
```diff
@@ -135,8 +135,14 @@ describe('optic-update lockForUpdate tests', function() {
                 {'update' : true, 'structure' : 'array'};
             db.rows.query(op.fromParam('bindingParam', null, outputCols).lockForUpdate(op.col('myUri')), options, {bindingParam: rows}).then((res) => {
                 try {
-                    res[0][0].name.should.equal('myUri');
-                    res[1][0].value.should.equal('/optic/test/musician1.json');
+                    if(serverConfiguration.serverVersion <= 11.1){
+                        res.columns[0].name.should.equal('myUri');
```
rjrudin, line 139: This suggests that if I ask for an "array" output type, I'm actually expecting an "object" output type. The "else" condition looks like it should be correct in both cases.

PR #844, opened 2024-02-15
MLE-12247 : Confirm Node client tests running on latest 10-nightly are passing.

anu3990:


test-complete/nodejs-temporal-protect-update.js
```diff
@@ -207,7 +207,7 @@ describe('Temporal protect update test', function() {
       response[0].metadataValues.temporalArchiveStatus.should.equal('succeeded,succeeded');
       response[0].metadataValues.temporalProtectExTime.should.equal(expTime);
       done();
-    }, done);
+    }).catch(error=>done(error));
   });
```
anu3990, line 211: I checked the response and the response only has -
metadataValues: { temporalDocURI: 'temporalUpdateDoc1.json' }
 Is there a way to ask for temporalArchivePaths, temporalProtectExTime etc in the response?

rjrudin: It's just doing a GET v1/documents and asking for metadataValues back, which means all metadata values should be retrieved. If temporalArchivePaths etc aren't there, then that likely means they don't exist on the document either. 

I found a Java Client test that calls "protect", and it does set some metadata fields I'd never heard of:


temporalProtectLevel | noDelete
temporalDocURI | javaSingleJSONDoc.json
temporalProtectExTime | 2024-02-16T12:24:05.519714Z

So check to see if the document has those metadata values on it. If not, check with the server team to see if they were removed in MarkLogic 10 nightly.

rjrudin (approved):
Approving so we can get this into the develop branch to see the regression failures there. My hunch is that the Node Client is making assertions on temporal-specific metadata fields that are not part of the public API for the bitemporal feature.

PR #847, opened 2024-03-14
Fix the validateDoc tests for MarkLogic 11.2+

BillFarber:


etc/test-lib.js
```diff
@@ -16,7 +16,7 @@
 'use strict';
 var valcheck = require('core-util-is');
 
-var mlutil    = require('../lib/mlutil.js');
+var mlutil = require('../lib/mlutil.js');
```
rjrudin, line 19: One note - try to keep whitespace changes like this in a separate commit/PR, as it makes the actual changes much more difficult to see - not just now, but in the future when someone looks at this commit to see what was fixed.

rjrudin (approved):
Looks good, don't worry about the whitespace changes in this as it's only changing a couple test files, just something to keep in mind for the future.

PR #849, opened 2024-03-22
Use ml-gradle to add an SSL server.

BillFarber:
Probably going to be multiple PRs for this story since the errors and fixes are probably unrelated.


Jenkinsfile
```diff
@@ -48,6 +48,10 @@ def runE2ETests(String type,String version){
         sleep 10
         node config-optic/setupqa.js
         sleep 30
+        cd test-complete-app
+        ./gradlew -i mlDeploy
+        cd ..
+        sleep 10
```
rjrudin, line 54: Why the sleep here? Any change made by mlDeploy is immediately committed by MarkLogic.

test-complete-app/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+  id "com.marklogic.ml-gradle" version "4.7.0"
+}
+
+/*
+ * Since our REST API server uses a certificate template, we need to execute an extra command to generate a temporary
+ * certificate for this template - without that, the REST API server won't be able to receive HTTP or HTTPS requests.
+ */
+ext {
+    def command = new com.marklogic.appdeployer.command.security.GenerateTemporaryCertificateCommand()
+    command.setTemplateIdOrName("ssl-project-template")
```
rjrudin, line 12: Use a different name for the template, like `mlxprs-ssl-template`, so it's obvious if someone is deploying this and other project's test apps to the same cluster what this is intended for.

test-complete-app/gradle.properties
```diff
@@ -0,0 +1,13 @@
+mlHost=localhost
+mlUsername=admin
+mlPassword=admin
+
+mlNoRestServer=true
+
+# Since the REST API server for this sample project is configured to use a certificate template for SSL, we set this
```
rjrudin, line 7: These comments don't look needed here. Even if `mlSimpleSsl` is needed at some point, we don't need all the comments copy/pasted here from an ml-gradle example project.

rjrudin (changes requested):

rjrudin (approved):

PR #854, opened 2024-03-27
Update Optic generated code.

BillFarber:


anu3990 (approved):

rjrudin (changes requested):
Take a look at `NewOpticMethodsInElevenDotOneTest` in the Java Client tests. I created that after running the code generator before 11.1 was released. I looked at the generated code and the generated tests, and I added those tests based on where the generated tests didn't provide any coverage of the new code. We'll want the same approach for every release of the Node/Java clients, as we can't assume that the generated code works correctly (though we do assume that the generated tests are reliable).

BillFarber:
Closing this until tests are written for then newly generated code.

PR #859, opened 2024-04-09
Fix the transaction time limit test

BillFarber:
Mocha does not guarantee "it" blocks are run sequentially or that they are non-synchronous.
Therefore, this "set" of tests really needs to be a single test with a chain of promises in order to guarantee order and timing.

### DIFF
--- a/test-complete/nodejs-documents-transaction-timelimit.js
+++ b/test-complete/nodejs-documents-transaction-timelimit.js
@@ -20,78 +20,103 @@ var testconfig = require('../etc/test-config-qa.js');
 var marklogic = require('../');
 
 var db = marklogic.createDatabaseClient(testconfig.restWriterConnection);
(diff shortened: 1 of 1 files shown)


test-complete/nodejs-documents-transaction-timelimit.js
```diff
@@ -20,78 +20,103 @@ var testconfig = require('../etc/test-config-qa.js');
...
+                            done(new Error('The read should fail since the transaction should have timed out.'));
+                        },
+                        function (err) {
+                            err.statusCode.should.equal(400);
+                            err.body.errorResponse.messageCode.should.equal('XDMP-NOTXN');
+                        }
+                    );
+            })
+            .then(() => {
+                // Attempt to check the transaction status again (it should fail)
```
rjrudin, line 109: Good thing "(it should fail)" is here.

PR #861, opened 2024-04-09
MLE-13686 : Use patchBuilder in Optic plan in Node Client

anu3990:


test-basic/optic-patchBuilder.js
```diff
@@ -0,0 +1,75 @@
...
+            const docColIdentifier = {uri: 'uri', doc:'doc', collections:'collections', metadata:'metadata', permissions:'permissions', quality:'quality', temporalCollection:'temporalCollection'}
+            db.rows.query(p.fromDocUris(p.cts.documentQuery('/graphQL-Publisher_1.xml'))
+                .joinDocCols(docColIdentifier, p.fragmentIdCol('fragmentId'))
+                .orderBy('uri')
+                .select(['uri', 'doc', 'metadata'])
+                .patch(p.col('doc'),
+                    p.patchBuilder('/')
+                        .insertAfter('/Citation', p.xmlComment('This is Test 01 comment after city element'))
+                ))
+                .then(function (response) {
```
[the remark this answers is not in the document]
BillFarber: That implies to me that somehow, the serialized query is not being generated correctly when you include the .insertAfter() call.
Have you looked at what is being generated? Rob's hint to me was to use the exportAs() method to look at the serialized query. Then, you can test it in QConsole.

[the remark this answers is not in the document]
rjrudin: Right - the way to debug this is to export the plan in the Node Client and then compare it to the result of exporting the same plan in qconsole.

lib/plan-builder-generated.js
```diff
@@ -6942,6 +6990,154 @@ bindParam(...args) {
...
+    * @param { Item } [node] - The path to insert the node before.
+    * @returns { planBuilder.PatchBuilder }
+    */
+insertBefore(...args) {
+    const namer = bldrbase.getNamer(args, 'path');
+    const paramdefs = [['path', [types.XsString], true, false], ['node', [types.Item, PlanColumn, PlanParam], false, true]];
+    const checkedArgs = (namer !== null) ?
+        bldrbase.makeNamedArgs(namer, 'PlanPatchBuilder.insertBefore', 2, new Set(['path', 'node']), paramdefs, args) :
+        bldrbase.makePositionalArgs('PlanPatchBuilder.insertBefore', 2, false, paramdefs, args);
+    return new PlanPatchBuilder('op', 'insert-before', checkedArgs);
```
rjrudin, line 7038: I think the problem may be here - when a new PlanPatchBuilder is created, you need to retain all of the arguments on the current PlanPatchBuilder.

The Java Client does this via the private `addArg` method in `PatchBuilderImpl`. That builds a new list of args, using the existing list of args. It then calls the private constructor on lines 30-32 which creates a new "op/suboperators" object and adds all the args that are passed in.

lib/plan-builder-generated.js
```diff
@@ -6942,6 +6990,154 @@ bindParam(...args) {
...
     }
 }
+/**
+  * PatchBuilder objects have methods.
+  * @namespace planBuilder.PatchBuilder
+  * @since 3.4.0
+  */
+class PlanPatchBuilder extends types.ServerType {
+
+  constructor(ns, fn, args) {
```
rjrudin, line 7000: I think this needs the same 3 constructors that `PatchBuilderImpl` has:

1. One that accepts just a context path.
2. One that accepts a context path and a map of namespaces.
3. A private one that accepts a list of args. 

Each constructor should then always build a new "op/suboperators" object and add args to it.

lib/plan-builder-generated.js
```diff
@@ -19,6 +19,7 @@
 
 const types    = require('./server-types-generated.js');
 const bldrbase = require('./plan-builder-base.js');
+let patchArgs = [];
```
rjrudin, line 22: I'm wondering if there's a thread safety issue here. If the user has a middle tier that uses the Node Client and two separate threads are building a plan, do they get their own instance of patchArgs, or are they trying to use the same instance of patchArgs? If it's the latter, they'd run into a thread safety issue as both threads would be populating this array, likely causing both plans to break.

lib/plan-builder-generated.js
```diff
@@ -6942,6 +6991,154 @@ bindParam(...args) {
 
     }
 }
+/**
+  * PatchBuilder objects have methods.
+  * @namespace planBuilder.PatchBuilder
+  * @since 3.4.0
+  */
+class PlanPatchBuilder extends types.ServerType {
+
```
rjrudin, line 7000: Oh hey - could `patchArgs` just be an attribute of this class instead? I think that would solve the thread safety issue - each thread would instantiate its own `PlanPatchBuilder` and therefore get its own `patchArgs` array.

rjrudin (changes requested):
Looks good, but try making `patchArgs` a field in `PlanPatchBuilder`, which I think should avoid any thread safety issues.

Also, add a comment to each part of the generated code that has been modified. That will make it much easier to ensure we don't lose the code the next time we run the generator.

lib/plan-builder-generated.js
anu3990: Tried that earlier but this or any other class in PlanBuilder do not allow variables. (If i try to add let patchArgs = [] on line 7000 it throws a syntax error). I don't think thread safety should be an issue here. In the test file, we are initializing PlanBuilder just once on line 18 and all the tests are using it. We can consider the tests as different threads here. They are accessing the same class but each have their own patchArgs and the queries are created accordingly.

rjrudin (approved):

PR #863, opened 2024-04-22
Security vulnerability fix and update for CHANGELOG.md

anu3990:


CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
 # CHANGELOG
+## 3.4.0
+#### New Functionality
+
+- Can now talk to MarkLogic using OAuth token (MarkLogic Server version 11.2.0 or higher needed).
```
rjrudin, line 5: Use "connect" here instead of "talk".

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
...
+- Can now submit optic update plans to /v1/rows/update (MarkLogic Server version 11.2.0 or higher needed).
```
rjrudin, line 6: Capitalize "Optic Update". And given that we don't have any other documentation for this, I think a brief mention of how this is accomplished is worth including. The user doesn't actually need to know about the "v1/rows/update" endpoint. 

Instead, how about something like: "Optic Update plans, including DSL queries that perform updates, now require `update: true` to be included in the options sent to the `rows.execute()` function."

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
...
+- Can now use PatchBuilder for patching documents (MarkLogic Server version 11.2.0 or higher needed).
```
rjrudin, line 7: A user won't know what "PatchBuilder" refers to here, we need to give them some context. So: "Can now perform partial updates on documents via Optic using the new `patchBuilder` operator." And I always try to include code mentions in backwards tick marks so they're formatted in Markdown, making them stand out more for the reader.

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
...
+- Can now use onError (MarkLogic Server version 11.2.0 or higher needed).
```
rjrudin, line 8: Same thing here, need to give the user context: "Can now capture errors from executing an Optic plan by including the new `onError` operator."

CHANGELOG.md
```diff
@@ -1,4 +1,16 @@
...
+
+#### Bug Fix
+
+- Can now send bindings to optic sever functions in the form of text.
```
rjrudin, line 12: Capitalize "Optic", and change "sever" to "server".

rjrudin (changes requested):

CHANGELOG.md
```diff
@@ -1,4 +1,17 @@
 # CHANGELOG
+## 3.4.0
+#### New Functionality
+
+- Can now connect to MarkLogic using OAuth token (MarkLogic Server version 11.2.0 or higher needed).
+- Optic Update plans, including DSL queries that perform updates, now require `update: true` to be included in the options sent to the `rows.execute()` function.
+     (MarkLogic Server version 11.2.0 or higher needed).
```
rjrudin, line 7: Just a small tweak, put this on the same line as above.

rjrudin (approved):
Looks good, just one small tweak and then merge away.

PR #866, opened 2024-04-25
Adding folders to npmignore to avoid publish to npm

anu3990:


rjrudin (changes requested):
Looks good, though add the following as well - CONTRIBUTING.md and Jenkinsfile. Users won't care about either of those.

rjrudin (approved):

PR #870, opened 2024-05-10
MLE-14123 : Jenkins pipeline fails due to gradle lock

anu3990:


test-app/gradle.properties
```diff
@@ -4,4 +4,5 @@ mlHost=localhost
 mlUsername=admin
 mlPassword=admin
 mlRestPort=8015
-mlContentForestsPerHost=3
\ No newline at end of file
+mlContentForestsPerHost=3
+org.gradle.daemon=false
```
rjrudin, line 8: Toss a comment on explaining why this is here - it's the kind of thing someone might look at a year from now and wonder "Huh, why is this here? I bet I can remove it." Comment just needs to explain that since the Jenkinsfile runs regression tests in parallel, we don't want Gradle to run in daemon mode.

rjrudin (approved):

PR #876, opened 2024-07-02
MLE-15292 : Find and remove reference for npm polyfill library in Node Client code.

anu3990:


package.json
```diff
@@ -81,6 +81,11 @@
     "nanoid": "3.1.31",
     "glob-parent": "6.0.2",
     "simple-get": "4.0.1",
-    "semver": "7.5.3"
+    "semver": "7.5.3",
+    "homedir-polyfill": "../_EXCLUDED_",
+    "inflight": "../_EXCLUDED_",
+    "rimraf": "5.0.7",
+    "@humanwhocodes/config-array": "@eslint/config-array",
+    "glob": "10.4.2"
```
rjrudin, line 89: I did a search and didn't find any references to rimraf or config-array - how do you know they're needed? "glob" is difficult to search for - it hits on "global" - but I'm likewise wondering how we know this needs to be added.

rjrudin (commented):
Generally, for any exclusions, we need to be very explicit in describing why something is being excluded. And for new inclusions, we want to document - ideally in the commit comment - why those dependencies are being included.

package.json
anu3990: When we import the dependencies, npm identifies the deprecated ones and the ones no longer supported. The replacement was also suggested by npm.

rjrudin: But do we actually need these? Unused dependencies only get more difficult to remove over time as it becomes more and more difficult to know if a dependency is needed or not. 

For example, I can't find any references to the string `rimraf` except in `package-lock.json`. That seems like a strong indicator that it's not needed. Maybe it still is, but I would opt for not including it and see if the Jenkins pipeline passes first. Same applies for `config-array`.

anu3990: config-array can be excluded but excluding rimraf throws error.

package.json
```diff
@@ -81,6 +81,11 @@
     "nanoid": "3.1.31",
     "glob-parent": "6.0.2",
     "simple-get": "4.0.1",
-    "semver": "7.5.3"
+    "semver": "7.5.3",
+    "homedir-polyfill": "../_EXCLUDED_",
```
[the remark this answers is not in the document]
anu3990: overrides section is mostly for transitive dependencies. There is more than one direct dependency that brings in homedir-polyfill. "npm list --depth=10" will display everything. Regarding Contrib file, since it has setup instructions, I don't think it is the right place to add this information. I have updated the jira attached to this PR with the reason why we are making this change.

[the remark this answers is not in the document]
rjrudin: Please add the reasons to the commit comments themselves. Given that this is an open source project, but not all users have access to JIRA, we want these exclusions to be publicly documented in the codebase itself. And unless there's a way to do that in package.json, the commit comments are the best place to very clearly explain why each exclusion exists - and also why each dependency is being added here that wasn't added before (my assumption being that they need to be added since they were previously brought in via a now-excluded dependency).

rjrudin (changes requested):

anu3990:
Added to commit message.

rjrudin (approved):

PR #879, opened 2024-07-09
MLE-15255 : Run the autogeneration of new Optic Code for Node Client

anu3990:


lib/plan-builder-generated.js
```diff
@@ -1194,8 +1194,7 @@ pointLongitude(...args) {
     * @returns { CtsPolygon }
     */
 polygon(...args) {
-    // Adding types.XsString manually since cts.polygon now accepts string input (MarkLogic 11.1.0 or higher needed).
-    const paramdef = ['vertices', [types.CtsPoint, PlanColumn, PlanParam, types.XsString], false, true];
+    const paramdef = ['vertices', [types.CtsPoint, PlanColumn, PlanParam], false, true];
     const checkedArgs = bldrbase.makeSingleArgs('cts.polygon', 1, paramdef, args);
```
anu3990, line 1198: Hi Aaron and Ling, the changes on lines 1197 and 1161 (cts.point and cts.polygon) came in as part of auto-generation but the tests are now failing. Please let me know if we can disregard the incoming changes from the auto-generation and revert it back.

rjrudin: Should definitely revert those changes. This is a good example of a comment added to a generated file helping us understand why it should be reverted.

lib/plan-builder-generated.js
```diff
@@ -8540,7 +8760,7 @@ fromDocDescriptors(...args) {
 
     }
 /**
-    * This function matches and returns the uri, content, and score for documents. Provides a client interface to a server function. See {@link http://docs.marklogic.com/op.fromDocUris|op.fromDocUris}
+    * This function matches and returns the uri for documents. Provides a client interface to a server function. See {@link http://docs.marklogic.com/op.fromDocUris|op.fromDocUris}
```
rjrudin, line 8763: This is fine for now, but in the future, any uses of "uri" (unless referring to a column name) should be changed in the hint/codegen file to be "URI".

rjrudin (commented):
Looks good, but definitely revert those changes to point/polygon, as that's undoing important additions for 11.1.0.

anu3990:
Moving ahead with https://github.com/marklogic/node-client-api/pull/880

PR #880, opened 2024-07-09
MLE-15255 : Run the autogeneration of new Optic Code for Node Client

anu3990:
(This generation was conducted after replacing hints file with the one in develop branch.)

### DIFF
--- a/lib/plan-builder-generated.js
+++ b/lib/plan-builder-generated.js
@@ -5300,6 +5300,220 @@ yearday(...args) {
     return new types.XsInteger('sql', 'yearday', checkedArgs);
     }
 }
+class VecExpr {
+  constructor() {
+  }
+  /**
+    *  Provides a client interface to a server function.
+    * @method planBuilder.vec#add
+    * @since 3.5.0
+    * @param { VecVector } [vector1] - The first addend vector.
+    * @param { VecVector } [vector2] - The second addend vector.
+    * @returns { VecVector }
+    */
+add(...args) {
+    const namer = bldrbase.getNamer(args, 'vector1');
+    const paramdefs = [['vector1', [types.VecVector, PlanColumn, PlanParam], true, false], ['vector2', [types.VecVector, PlanColumn, PlanParam], true, false]];
+    const checkedArgs = (namer !== null) ?
+        bldrbase.makeNamedArgs(namer, 'vec.add', 2, new Set(['vector1', 'vector2']), paramdefs, args) :
+        bldrbase.makePositionalArgs('vec.add', 2, false, paramdefs, args);
+    return new types.VecVector('vec', 'add', checkedArgs);
+
+    }
+/**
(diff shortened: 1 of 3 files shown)


lib/plan-builder-generated.js
```diff
@@ -1155,10 +1155,9 @@ periodRangeQuery(...args) {
 point(...args) {
     const namer = bldrbase.getNamer(args, 'latitude');
     const paramdefs = [['latitude', [types.XsDouble, PlanColumn, PlanParam], true, false], ['longitude', [types.XsDouble, PlanColumn, PlanParam], true, false]];
-    // Changing minArity to 1 manually since cts.point now accepts string input (MarkLogic 11.1.0 or higher needed).
```
rjrudin, line 1158: Still need to undo these changes to cts.point and cts.polygon.

rjrudin (changes requested):

test-basic/plan-builder-generated.js
```diff
@@ -1732,6 +1744,36 @@ describe('plan builder', function() {
             done();
         }).catch(done);
     }); 
+    it('vec.base64Decode#1', function(done) {
+        if(serverConfiguration.serverVersion < 12) {
+            this.skip();
+        }
+        testPlan([p.xs.string("abc")], p.vec.base64Decode(p.col("1")))
```
rjrudin, line 1751: What is this test actually doing? It looks like it's making sure that the response has "null" in it somewhere - how do we know that's ensuring that base64Decode works? 

I think the tests that @BillFarber  added for the vector operations for the Java Client would be much better - i.e. create a plan that binds new columns with the output of a bunch of vector operations, and verify that the bound columns have the expected values.

test-basic/plan-builder-generated.js
```diff
@@ -2353,8 +2395,12 @@ describe('plan builder', function() {
     }); 
     it('xs.negativeInteger#1', function(done) {
         testPlan([p.xs.double(-1)], p.xs.negativeInteger(p.col("1")))
-          .then(function(response) { 
-            should(String(getResult(response).value).replace(/^ /, '')).equal('-1');
+          .then(function(response) {
+              if(serverConfiguration.serverVersion < 12) {
+                  should(String(getResult(response).value).replace(/^ /, '')).equal('-1');
+              } else {
+                  should(String(getResult(response).value).replace(/^ /, '')).equal(null);
```
rjrudin, line 2402: Why is there here for MarkLogic 12? Is it expected that this now returns null instead of -1?

test-basic/plan-builder-generated.js
anu3990: This was auto-generated. I added the skip block since it fails on server versions less than 12.

test-basic/plan-builder-generated.js
anu3990: Since it was auto-generated I would say yes, but for server versions less than 12, it gives -1 so added the if else block.

rjrudin (approved):
Approving in anticipation of manually-written tests being added for the vector operations.

llinggit (approved):

rjrudin (approved):

PR #889, opened 2024-07-19
MLE-4113 : Submit Optic fromSearch plan with BM25 scoring with Node Client

anu3990:


lib/plan-builder-base.js
```diff
@@ -122,11 +122,11 @@ function castArg(arg, funcName, paramName, argPos, paramTypes) {
         const value = arg[key];
         switch(key) {
           case 'scoreMethod':
-            if (['logtfidf', 'logtf', 'simple', 'score-logtfidf', 'score-logtf', 'score-simple'].includes(value)) {
+            if (['logtfidf', 'logtf', 'simple', 'score-logtfidf', 'score-logtf', 'score-simple', 'zero', 'random', 'bm25'].includes(value)) {
               return true;
             }
             throw new Error(
-                `${argLabel(funcName, paramName, argPos)} can only be logtfidf, logtf, or simple`
+                `${argLabel(funcName, paramName, argPos)} can only be logtfidf, logtf, simple, 'zero', 'random' or bm25`
```
rjrudin, line 129: I like the use of single quotes around the values, but it should be done for every value.

test-basic/plan-search.js
```diff
@@ -166,8 +178,76 @@ describe('search', function() {
...
+        done();
+      }
+    });
+
+    it('should search documents with zero scoring method', function(done) {
+      execPlan(
+          p.fromSearch(p.cts.jsonPropertyValueQuery('instrument', 'trumpet'),
+              ['score', 'quality'], null, { scoreMethod: 'zero'})
+      ).then(function(response) {
+        assert(response.columns != null)
```
rjrudin, line 236: This test is worth adding an assertion to verify that each row has a column named "score" with value of zero. That also helps ensure that `scoreMethod` is working regardless of the value.

rjrudin (changes requested):
Looks good, just a couple small tweaks to make.

lib/plan-builder-base.js
```diff
@@ -122,11 +122,11 @@ function castArg(arg, funcName, paramName, argPos, paramTypes) {
         const value = arg[key];
         switch(key) {
           case 'scoreMethod':
-            if (['logtfidf', 'logtf', 'simple', 'score-logtfidf', 'score-logtf', 'score-simple'].includes(value)) {
+            if (['logtfidf', 'logtf', 'simple', 'score-logtfidf', 'score-logtf', 'score-simple', 'zero', 'random', 'bm25'].includes(value)) {
               return true;
             }
             throw new Error(
-                `${argLabel(funcName, paramName, argPos)} can only be logtfidf, logtf, or simple`
+                `${argLabel(funcName, paramName, argPos)} can only be 'logtfidf', 'logtf', 'simple', 'zero', 'random' or 'bm25'`
```
rjrudin, line 129: Sorry, I should have noticed this initially - the values that are checked above and the values listed here are not the same. 

Based on what's in plan-search.js, looks like `simple` is valid but not `score-simple`. I'm assuming that the 3 values checked on line 125 that start with `score-` are not valid and should be removed. That will result in both lists having a length of 6. That also matches what's at https://docs.marklogic.com/cts:search , except that while the low-level function requires "score-" as a prefix, the REST API does not.

rjrudin (changes requested):

rjrudin (approved):

PR #903, opened 2024-09-11
MLE-16370 : Run optic auto-generation for op:shortest-path

anu3990:


rjrudin (changes requested):
Need a test for this.

anu3990:
We have a different jira for the overall implementation - https://progresssoftware.atlassian.net/browse/MLE-15256. Will add the tests in that.

PR #904, opened 2024-10-01
MLE-15256 : Use shortest-path function in Node plan builder

anu3990:


lib/plan-builder-generated.js
```diff
@@ -8024,6 +8024,26 @@ unnestLeftOuter(...args) {
         bldrbase.makePositionalArgs('PlanModifyPlan.unnestLeftOuter', 2, false, paramdefs, args);
     return new PlanModifyPlan(this, 'op', 'unnest-left-outer', checkedArgs);
 
+    }
+/**
+    * This method can be used to find the shortest path between two nodes in a given graph. Provides a client interface to a server function. See {@link http://docs.marklogic.com/ModifyPlan.prototype.shortestPath|ModifyPlan.prototype.shortestPath}
+    * @method planBuilder.ModifyPlan#shortestPath
+    * @since 3.6.0
+    * @param { PlanExprColName } [start] - The column representing the input starting subject of the shortest path search. The columns can be named with a string or a column parameter function such as op:col.
```
rjrudin, line 8032: Something to check with Ling on - for a Node or Java user, seeing "op:col" in the docs will be confusing as that's an XQuery function name and doesn't apply here. Even "op.col" will be a little confusing, as that's a server-side JavaScript function in the "op" namespace. Really, it would be better if the "such as op:col" part was removed from each of the descriptions as it's misleading, if not incorrect, information in our clients. 

It's fine for this PR, but check with Ling if that description can be removed so it doesn't show up here.

rjrudin (approved):

PR #907, opened 2024-11-25
MLE-17147 Added annTopK support

rjrudin:


lib/plan-builder-generated.js
```diff
@@ -7470,6 +7470,28 @@ class PlanModifyPlan extends PlanPreparePlan {
   constructor(prior, ns, fn, args) {
     super(prior, ns, fn, args);
   }
+
+  /**
```
BillFarber, line 7474: Since the top of the file says that the file shouldn't be edited, I wonder if we should add something to the comment (at the top) that says something like, "Except for certain additions, such as 'annTopK'"

BillFarber (dismissed):

lib/plan-builder-generated.js
rjrudin: I'll remove that comment as it'll never be possible to fully generate this file, too many issues occur.

BillFarber (approved):

PR #910, opened 2025-01-24
MLE-19469 : Drop "weight" as a parameter in shortest-path in Node Client

anu3990:


anu3990:
No changes needed for the existing shortest-path tests.

PR #919, opened 2025-04-09
MLE-16819 : Support TLSv1.3 via Node Client

anu3990:


lib/marklogic.js
```diff
@@ -793,6 +793,9 @@ function initClient(client, inputParams) {
       mlutil.copyProperties(inputParams, agentOptions, [
         'keepAliveMsecs', 'maxCachedSessions', 'maxFreeSockets', 'maxSockets', 'maxTotalSockets', 'scheduling', 'timeout'
       ]);
+      agentOptions.secureProtocol = 'TLS_method';
+      agentOptions.secureOptions = require('constants').SSL_OP_NO_TLSv1 | require('constants').SSL_OP_NO_TLSv1_1;
```
BillFarber, line 797: Is there a reason you're disabling these versions of TLS in the client? I don't think that was the intention of the story.

test-basic/ssl-min-allow-tls-test.js
```diff
@@ -0,0 +1,85 @@
...
+            contentType: 'application/json',
+            content: '{"key1":"With TLS 1.3"}'
+        })
+            .result(function (response) {
+                db.documents.read('/test/write_tlsV1.3.json')
+                    .result(function (documents) {
+                        documents[0].content.should.have.property('key1');
+                        documents[0].content.key1.should.equal('With TLS 1.3');
+
+                    }).then(()=>{
```
BillFarber, line 49: I think readability and maintainability would be improved here if these were different tests.

test-basic/ssl-min-allow-tls-test.js
```diff
@@ -0,0 +1,85 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+let testconfig = require('../etc/test-config.js');
+let should = require('should');
+let marklogic = require('../');
+const { exec } = require('child_process');
+let db = marklogic.createDatabaseClient(testconfig.restWriterConnection);
```
BillFarber, line 21: We need two clients here. One with TLSv1.3 enabled (use this in the existing tests), and one with TLSv1.2 enabled.

Then there should be another test with an attempt to use the TLSv1.2 client with a DB set to minimum TLSv1.3, and expect an exception.

BillFarber (changes requested):

lib/marklogic.js
anu3990: So the default will be 1.3 and the fall back will be 1.2. Disabling version 1 and version 1.1 will make sure the fall back version will be 1.2.

test-basic/ssl-min-allow-tls-test.js
anu3990: Will do. Also, I will include option to run the tests only for MarkLogic 12.

anu3990:
Updated this PR. Tests are running successfully on 12-nightly.

rjrudin (changes requested):

rjrudin (approved):

lib/marklogic.js
anu3990: These changes are no longer there and as Rob suggested, min and max versions were included.

PR #922, opened 2025-04-21
MLE-17209 : Review and fix vulnerabilities in Polaris report for Node Client.

anu3990:


rjrudin (approved):
Looks good! I like the simple use of `!` instead of checking null or undefined.

PR #926, opened 2025-05-07
Adding "export default" in two transforms

rjrudin:
Required for these to work on MarkLogic 12 nightly.


rjrudin:
Closing this for now. Jenkins isn't running against the latest 12 nightly, so this actually caused the tests to fail.

PR #927, opened 2025-05-14
MLE-21460 : ResultProvider.result() sometimes returns a text node as a Buffer

anu3990:


lib/responder.js
```diff
@@ -1119,6 +1119,11 @@ function resolvedPromise(operation, resolve) {
   }
 
   var data    = operation.data;
+  for(let i=0; i<data.length; i++){
+    if(data[i].format === 'text' && typeof data[i].value !== 'string'){
+      data[i].value = data[i].value.toString();
+    }
+  }
   var hasData = (data != null);
```
rjrudin, line 1127: Given this null check here, I think you need to add checks before your code to make sure data is not null and that data has `length` as well.

lib/responder.js
```diff
@@ -1119,6 +1119,11 @@ function resolvedPromise(operation, resolve) {
   }
 
   var data    = operation.data;
+  for(let i=0; i<data.length; i++){
+    if(data[i].format === 'text' && typeof data[i].value !== 'string'){
```
rjrudin, line 1123: For sanity purposes, include a null check on `data[i].value` as well as we don't want to call `toString()` on it if it's null.

rjrudin (changes requested):

rjrudin (approved):

PR #930, opened 2025-05-28
MLE-21839 : test-teardown.js throws error on Jenkins

anu3990:


rjrudin (commented):
I actually think the solution here is to get rid of test-teardown.js. There's no reason for the test suite to be deleting anything that was setup via mlDeploy on the test-app folder. The expectation after the tests are run is that the test app is still present. It should only be removed via an mlUndeploy call - but there's no reason for the test suite to do that either.

anu3990:
Let me check with Sameera if we can remove the teardown step but if it is needed, we will need to fix the error.

anu3990:
Looks like Jenkins is running fine after removing the teardown step. We can proceed with this. Will come back to this if there are issues later. Thanks.!

rjrudin (approved):

PR #936, opened 2025-07-02
MLE-20906 : Update Node Client based on changes to functions in Optic

anu3990:


rjrudin (approved):

anu3990:
Updated vectorScore tests as suggested by Lazaro.

PR #954, opened 2025-08-15
MLE-23402 : MarkLogic Node.js Client API's graph was changed to graph_, rendering documentation links hard to maintain.

anu3990:


lib/graphs.js
```diff
@@ -86,6 +86,9 @@ function listOutputTransform(headers, data) {
   return data;
 }
 
+/**
```
rjrudin, line 89: If this does the trick, add a comment here explaining why this annotation is here, as a future developer won't know why it's here and may remove it, not knowing what it does.

rjrudin (approved):

PR #962, opened 2025-09-18
Remove www-authenticate dependency

### LINKED ISSUE #961: The `www-authenticate` library used for DIGEST authentication causes runtime crash in FIPS mode
Hi all!

We're helping a client get MarkLogic to use certificate-based authentication under FIPS mode. We've got the authentication working without FIPS, but once enabled the `www-authenticate` library used by `marklogic` immediately crashes the runtime in FIPS mode. The `www-authenticate` package (https://www.npmjs.com/package/www-authenticate) is abandoned, so we should not expect a change there. The issue there is quite simple:

https://github.com/randymized/www-authenticate/blob/master/lib/md5.js#L2 immediately creates a md5sum at require time (presumably as a functionality test) 

This will crash under FIPS even if DIGEST auth is not being used, simply due to `www-authenticate` package being present at all. Simply removing that line is enough to resolve the issue under FIPS.

I was not able to immediately find a quick replacement package. As a potential solution I've put together a zero-dependency DIGEST implementation here: https://github.com/marklogic/node-client-api/pull/962. Let me know if you have alternative approaches, but this is a blocker for our client.

brycebaril:
- Added custom digest-auth.js module supporting RFC 7616 and RFC 2069
- Removed www-authenticate package dependency
- Updated requester.js to use digest-auth.js
- Added unit tests that should cover all MarkLogic scenarios

Fixes #961

### DIFF
--- a/lib/digest-auth.js
+++ b/lib/digest-auth.js
@@ -1,0 +1,135 @@
+'use strict';
+
+// Zero-dependency HTTP Digest Access Authentication helper.
+// Covers RFC 7616 MD5 and MD5-sess algorithms with qop="auth" (default).
+// Falls back to the legacy RFC 2069 variant if the server omits qop.
+//
+// Usage:
+//   const createDigestAuth = require('./digest-auth');
+//   const auth = createDigestAuth(user, pass, challengeHeader);
+//   request.setHeader('Authorization', auth.authorize(method, path, bodyOptional));
+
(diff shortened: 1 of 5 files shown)


rjrudin:
Thanks @brycebaril ! We've got our hands full with a few things right now but will try to review this soon. Thanks for putting together both the issue and the PR.

PR #1045, opened 2025-12-19
MLE-25586 add trans closure to node client

stevebio:
Add transitive closure function to node client. Function body generated by Optic Code generator. Manual changes to base, and creation of tests. Increase timeout on an SSL v1.2 test, passes most of the time now Skip a broken test, will fix later.

### DIFF
--- a/lib/plan-builder-base.js
+++ b/lib/plan-builder-base.js
@@ -401,6 +401,26 @@ function castArg(arg, funcName, paramName, argPos, paramTypes) {
             });
           }
           return true;
+        case 'PlanTransitiveClosureOptions':
+          const planTransitiveClosureOptionsSet = new Set(['minLength', 'min-length', 'maxLength', 'max-length']);
+          if(Object.getPrototypeOf(arg) === Map.prototype){
+            arg.forEach((value, key) => {
+              if(!planTransitiveClosureOptionsSet.has(key)) {
+                throw new Error(
+                    `${argLabel(funcName, paramName, argPos)} has invalid key- ${key}`
+                );
+              }
+            });
+          } else if (typeof arg === 'object') {
+            Object.keys(arg).forEach(key => {
+              if(!planTransitiveClosureOptionsSet.has(key)) {
+                throw new Error(
+                    `${argLabel(funcName, paramName, argPos)} has invalid key- ${key}`
+                );
+              }
+            });
+          }
+          return true;
       default:
         return false;
       }
--- a/test-basic/service-caller.js
+++ b/test-basic/service-caller.js
@@ -70,7 +70,8 @@ describe('Service caller', function() {
             });
     });
 
-    it('postOfUrlencodedForDocumentArray1 endpoint', function(done) {
+    // errors all the time now, should fix.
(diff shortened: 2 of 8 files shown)


lib/plan-builder-base.js
```diff
@@ -401,6 +401,26 @@ function castArg(arg, funcName, paramName, argPos, paramTypes) {
             });
           }
           return true;
+        case 'PlanTransitiveClosureOptions':
+          const planTransitiveClosureOptionsSet = new Set(['minLength', 'min-length', 'maxLength', 'max-length']);
```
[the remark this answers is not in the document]
stevebio: This is consistent with previous impl

test-basic/service-caller.js
```diff
@@ -70,7 +70,8 @@ describe('Service caller', function() {
             });
     });
 
-    it('postOfUrlencodedForDocumentArray1 endpoint', function(done) {
+    // errors all the time now, should fix.
```
rjrudin, line 73: This is passing consistently on Jenkins - https://ml-clt-jenkins.progress.com/job/devexp/job/Node-Client/job/Node-client-api/job/develop/160/ - so maybe just a local issue? Does it fail when you run all of "test-basic"?

test-basic/ssl-min-allow-tls-test.js
```diff
@@ -12,7 +12,7 @@ let serverConfiguration = {};
 let host  = testconfig.testHost;
 
 describe('document write and read using min tls', function () {
-    this.timeout(10000);
+    this.timeout(12000);
```
rjrudin, line 15: Was this failing for you locally too? It's another test that I don't think has failed on Jenkins in quite a while. Of course, 10s is already totally arbitrary, so changing it to 12s doesn't really hurt.

test-basic/transitive-closure.js
```diff
@@ -0,0 +1,141 @@
...
+                done();
+            }, 3000);
+        } catch (error) {
+            done(error);
+        }
+    });
+
+    it('with simple pattern full transitive closure', function (done) {
+        execPlan(
+            p.fromTriples([
```
rjrudin, line 42: This is great, just getting a test case in place is like 98% of the value of the PRs for these new Optic functions, as now we have a way to test / demo it. I'll try it out locally.

rjrudin (commented):
About to test locally, will respond back.

test-basic/ssl-min-allow-tls-test.js
stevebio: Yes the timing for the error ssl v1.2 error test was like 11000ms for me. so 120000 timeout fixed locally. But totally environmental and not even close to a good solution. I could bump to 200000.

test-basic/transitive-closure.js
```diff
@@ -0,0 +1,141 @@
+/*
```
rjrudin, line 1: A few of the imports in here are flagged by VS Code as unused (I guess the `nom run lint` doesn't complain on unused imports yet) - should, assert, and restWriterConnection. Remove those and then merge away!

rjrudin (approved):

PR #1046, opened 2025-12-30
MLE-26340 expose vec.precision and vec.trunc in Node client.

stevebio:


lib/plan-builder-base.js
```diff
@@ -59,7 +59,7 @@ function castArg(arg, funcName, paramName, argPos, paramTypes) {
       arg = arg.valueOf();
     } else if (arg instanceof types.ServerType) {
       if(arg._ns === 'vec'){
-        return arg._args;
+        return arg;
```
rjrudin: Whatever the right line of code is here, we definitely need a comment explaining why it's okay not to throw an error if the namespace is "vec", and then why the arg should actually be the value of "_args" (the previous code) or why it doesn't need to be cast at all. This "if ServerType" block is particularly confusing for a new reader (me) because there's no explanation too as to why an Error was being thrown before, which presumably would only have been in the case of programmer error.

test-basic/optic-vector.js
```diff
@@ -34,129 +34,191 @@ describe('tests for new vector functions.', function() {
...
-        const vec2 = p.vec.vector([0.000000000001, 0.000000000002]);
-        testPlan([""],p.vec.add(p.vec.subvector(vec1,0),p.vec.subvector(vec2,1)))
+        const vec1 = op.vec.vector([1]);
+        const vec2 = op.vec.vector([2]);
+        testPlan([""],op.vec.add(vec1, vec2))
             .then(function(response) {
-               assert(response.rows[0].t.value[0][0] =='1e-12')
-               assert(response.rows[0].t.value[1][0]=='2e-12')
+               // add([1], [2]) = [3] (element-wise addition)
+               assert(response.rows[0].t.value[0] == 3, 'vec.add did not return expected value: [1] + [2] should be [3]');
```
rjrudin: The assertion message should remove the need for the comment above - i.e. they're both explaining the same thing. Just use an assertion message.

lib/plan-builder-generated.js
```diff
@@ -5458,6 +5458,23 @@ normalize(...args) {
     const paramdef = ['vector1', [types.VecVector, PlanColumn, PlanParam], false, false];
     const checkedArgs = bldrbase.makeSingleArgs('vec.normalize', 1, paramdef, args);
     return new types.VecVector('vec', 'normalize', checkedArgs);
+    }
+/**
+    * Returns a new vector which is a copy of the input vector with reduced precision. The precision reduction is achieved by clearing the bottom (32 - precision) bits of the mantissa for each dimension's float value. This can be useful for reducing storage requirements or for creating approximate vector representations. Provides a client interface to a server function. See {@link http://docs.marklogic.com/vec.precision|vec.precision}
+    * @method planBuilder.vec#precision
+    * @since 4.1.0
+    * @param { VecVector } [vector] - The input vector to reduce precision.
+    * @param { XsUnsignedInt } [precision] - The number of mantissa bits to preserve (9-32 inclusive). Default is 16. Higher values preserve more precision. If the value is outside the valid range, throw VEC-INVALIDPRECISION.
```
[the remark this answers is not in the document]
stevebio: JSDoc is auto-generated

stevebio:
Ling added vec.precision and vec.trunc and now op calls work.

test-basic/optic-vector.js
```diff
@@ -1,12 +1,12 @@
...
-* Copyright (c) 2015-2025 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
+* Copyright (c) 2015-2026 Progress Software Corporation and/or its subsidiaries or affiliates. All Rights Reserved.
 */
 'use strict';
 
 const should = require('should');
 
 const marklogic = require('../');
-const p = marklogic.planBuilder;
+const op = marklogic.planBuilder;
```
rjrudin, line 9: ❤️

rjrudin (approved):
Looks good!

PR #1049, opened 2026-01-07
MLE-23744 implement optic fromDocs in node client

stevebio:
Add tweaked auto-generated functions and classes for optic fromDocs plan builder. Modify PlanColumnBuilder by adding PlanColumnBuilderBase to allow chaining during column definitions. This follows the patch builder pattern. Move transitiveClosure function after annTopK to match auto-generated code and make diff easier.

### DIFF
--- a/lib/plan-builder-generated.js
+++ b/lib/plan-builder-generated.js
@@ -14,6 +14,7 @@ version numbers for "since" annotations.
 const types    = require('./server-types-generated.js');
 const bldrbase = require('./plan-builder-base.js');
 let patchArgs = [];
+let columnBuilderArgs = [];
 
 class CtsExpr {
   constructor() {
@@ -5532,11 +5533,11 @@ trunc(...args) {
     * Returns a vector value. Provides a client interface to a server function. See {@link http://docs.marklogic.com/vec.vector|vec.vector}
     * @method planBuilder.vec#vector
     * @since 3.5.0
-    * @param { XsAnyAtomicType } [values] - The value(s) to create the vector from. Can be a sequence or json:array of integer or floating-point numbers. Also accepts a string that has the format of a JSON array of Numbers. Also accepts a string that was created by vec:base64-encode().
+    * @param { XsAnyAtomicType } [values] - The value(s) to create the vector from. Can be a sequence or json:array of integer or floating-point numbers. Also accepts a string that has the format of a JSON array of Numbers, a string that was created by vec:base64-encode(), or a types.Node such as returned by xpath.
     * @returns { VecVector }
     */
 vector(...args) {
-    const paramdef = ['values', [types.XsAnyAtomicType, PlanColumn, PlanParam], false, true];
+    const paramdef = ['values', [types.XsAnyAtomicType, types.Node, PlanColumn, PlanParam], false, true];
     const checkedArgs = bldrbase.makeSingleArgs('vec.vector', 1, paramdef, args);
     return new types.VecVector('vec', 'vector', checkedArgs);
     }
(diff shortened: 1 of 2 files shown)


lib/plan-builder-generated.js
```diff
@@ -7135,12 +7136,167 @@ class PlanAnnTopKOptions extends types.ServerType {
...
+  */
+class PlanColumnBuilder extends types.ServerType {
+
+  constructor(ns, fn, args) {
+    super(ns, fn, args);
+  }
+/**
+    * Create a new column definition for use with op:from-docs. This method initializes the column with default properties unless overridden by additional chained methods. Default Behavior: If no other properties are set (such as xpath(), type(), or nullable()), the column defaults to: XPath: ./name (where name is the name passed to add-column()).Type: stringNullable: true Provides a client interface to a server function. See {@link http://docs.marklogic.com/op.addColumn|op.addColumn}
+    * @method planBuilder.ColumnBuilder#addColumn
+    * @since 2.1.1
```
rjrudin: Watch out when these old values get used somehow, probably just a find/replace to bump to 4.1.0.

lib/plan-builder-generated.js
```diff
@@ -7135,12 +7136,167 @@ class PlanAnnTopKOptions extends types.ServerType {
...
+    * @param { PlanColumnName } [column] - The Column Builder Plan. You can either use the XQuery =&gt; chaining operator or specify the variable that captures the return value from the previous operation.
```
rjrudin, line 7158: This isn't something we want to change (I think) but I just mention it as an example of where the code generator leaks server-side details into the client code. This mention of xquery is of course irrelevant for a client user.

test-basic/optic-fromDocs.js
```diff
@@ -0,0 +1,59 @@
...
+        }
+    });
+
+    describe('fromDocs', function () {
+
+        it('basic', function (done) {
+            const plan = op.fromDocs(
+                    op.cts.wordQuery('Coltrane'),
+                    '/musician',
+                    op.columnBuilder()
```
rjrudin, line 157: This looks like a great start to me - I think you should ping Jos too to just have him skim through this as a quick sanity check, as he's the fromDocs author.

lib/plan-builder-generated.js
stevebio: Eeks, fixing

stevebio: So more eye-crossing diff lines to add to the pile.

stevebio: These are all actually bogus, need real comments for these. I'll fix.

test-basic/optic-fromDocs.js
stevebio: will do, once I get a few more tests in place.

lib/plan-builder-generated.js
stevebio: Good catch.

stevebio:
One pattern breaker: I added "types.Node" as a new additional param type for the vector() function. This matches other functions like op.xs.decimal in that they can accept an xpath result node. This supports an optic pattern like this that seems vital for the node client within fromDocs:
...
op.vec.vector(op.xpath(op.context(), './embedding'))
...
This won't work in node unless types.Node is allowed to vector() function.
This pattern works in optic already.
Review the last test in optic-fromDocs and let me know if you think this is wrong.

rjrudin (approved):
Ran the tests locally on the latest nightly build, looks good! The tests do a great job of documenting the feature.

PR #1096, opened 2026-07-14
MLE-30263: Fix Uncaught Exception in getAccessToken Error Handler

ngodugu-marklogic:
**Jira:** [MLE-30263](https://progresssoftware.atlassian.net/browse/MLE-30263)
**Severity:** Medium | CVSS: 7.5 High | CWE: CWE-755

### Problem
In Cloud authentication (`authType: 'CLOUD'`), `getAccessToken()` threw errors from async event handlers (`res.on('data')` and `res.on('error')`). Throwing inside these callbacks can cause uncaught exceptions and crash the Node.js process instead of rejecting the calling operation.

The implementation also double-wrapped errors (`new Error(existingError)`), producing confusing error messages.

### Solution
Updated token retrieval error handling in `lib/requester.js` to use operation-level error propagation instead of throw-based async failures.

- Added internal `handleTokenError(error, contextMessage)` helper that:
  - ensures token failures are reported once (`tokenErrorReported`)
  - releases `operation.lockAccessToken` on failure
  - routes errors through `operation.errorListener(new Error(...))`
- Replaced `statusCode === 400` throw path with:
  - `handleTokenError(null, 'Token endpoint returned 400: ...')`
  - immediate return
- Added defensive token response parsing/validation:
  - wraps JSON parsing in `try/catch`
  - validates `access_token` and `.expires`
  - routes invalid payloads via `operation.errorListener`
- Replaced response error throw path with `handleTokenError(...)`.
- Added `req.on('error', ...)` handling for request/socket failures (e.g. `ECONNRESET`).

### Changes
File	Change
lib/requester.js	Reworked `getAccessToken()` error paths to use `operation.errorListener`; removed throw-based async failures; added request-error handling and token response validation
test-basic/cloud_authentication-test.js	Added 2 regression tests for request network errors and token endpoint 400 responses

### Tests
Pre-existing tests in `test-basic/cloud_authentication-test.js` continue to pass (with one existing skipped test).

New tests added:
- `should route token request network errors to operation.errorListener`
- `should route token endpoint 400 responses to operation.errorListener`

Validated result:
```
- 6 passing
- 1 pending
```

Testing instructions
```
Project requires Node >= 22
```

Install dependencies
```
npm ci --no-audit --no-fund
```

Run targeted cloud auth tests
```
npx mocha test-basic/cloud_authentication-test.js
```

If your default Node is not 22, run explicitly with Node 22
```
npx -y node@22 node_modules/mocha/bin/mocha.js test-basic/cloud_authentication-test.js
```

[MLE-30263]: https://progresssoftware.atlassian.net/browse/MLE-30263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/requester.js
+++ b/lib/requester.js
@@ -47,6 +47,18 @@ function getAuthenticatorKerberos(client) {
 function getAccessToken(operation){
   const postData = `${encodeURI('grant_type')}=${encodeURI('apikey')}&${encodeURI('key')}=${encodeURI(operation.client.connectionParams.apiKey)}`;
   const path = operation.client.connectionParams.accessTokenDuration?'/token?duration='+encodeURIComponent(operation.client.connectionParams.accessTokenDuration):'/token';
+  function handleTokenError(error, contextMessage) {
+    if (!operation.lockAccessToken) {
+      return;
+    }
+
+    operation.lockAccessToken = false;
+
+    const baseMessage = (contextMessage == null) ? 'Failed to obtain access token' : contextMessage;
+    const errorMessage = (error && error.message) ? `${baseMessage}: ${error.message}` : baseMessage;
+    operation.errorListener(new Error(errorMessage));
+  }
+
   const options = {
     hostname: operation.client.connectionParams.host,
     port: 443,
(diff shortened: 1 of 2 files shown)


lib/requester.js
```diff
@@ -58,26 +70,49 @@ function getAccessToken(operation){
...
       }
-      const responseValue = JSON.parse(d.toString());
-      operation.accessToken = responseValue.access_token;
-      operation.expiration = new Date(responseValue['.expires']);
-      if(operation.lockAccessToken){
+
+      try {
+        const responseValue = JSON.parse(responseBody);
+        if (!responseValue.access_token || !responseValue['.expires']) {
+          throw new Error('missing required token fields');
```
RitaChen609, line 88: throw new Error is used here. Should it be replaced with handleTokenError and return earlier?

lib/requester.js
```diff
@@ -58,26 +70,49 @@ function getAccessToken(operation){
...
+        }
+
+        const expiration = new Date(responseValue['.expires']);
+        if (Number.isNaN(expiration.getTime())) {
+          throw new Error('invalid .expires value');
```
RitaChen609, line 93: throw new Error is used here. Should it be replaced with handleTokenError?

rjdew-progress: I think that this is intentional to be picked up by `req.on('error', (e) => `

lib/requester.js
```diff
@@ -60,22 +77,34 @@ function getAccessToken(operation){
   const req = https.request(options, (res) => {
     res.on('data', (d) => {
       if(res.statusCode === 400){
-        throw new Error(d.toString());
+        handleTokenError(null, 'Token endpoint returned 400: ' + d.toString());
+        return;
```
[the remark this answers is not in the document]
rjdew-progress: This is in an `res.on('end', () => {` function block, not a `res.on('data', (d) => {` function block

rjdew-progress (dismissed):

lib/requester.js
RitaChen609: I think the throw new Error would be caught by catch(error) block (line 100) later. It then will call handleTokenError. So I think using throw new Error is not a bug, but replacing with handleTokenError and return earlier would be clearer and more explicit.

lib/requester.js
```diff
@@ -58,26 +70,49 @@ function getAccessToken(operation){
...
+
+        const expiration = new Date(responseValue['.expires']);
+        if (Number.isNaN(expiration.getTime())) {
+          throw new Error('invalid .expires value');
+        }
+
+        operation.accessToken = responseValue.access_token;
+        operation.expiration = expiration;
         operation.lockAccessToken = false;
+        authenticatedRequest(operation);
```
RitaChen609: If authenticatedRequest throws for any reason unrelated to token parsing, that error will be caught and reported as "Invalid token endpoint response", which is misleading. Should authenticatedRequest be moved outside the try block?

ngodugu-marklogic: Makes Sense.. Fixed it!

RitaChen609 (approved):

jonmille (approved):

PR #1106, opened 2026-08-10
MLE-30272: [Node Client] Document ML-Agent-ID Telemetry Header and Provide Opt-Out Option

ngodugu-marklogic:
### Summary
This PR documents the ML-Agent-ID telemetry header behavior and adds an opt-out connection option so security-sensitive deployments can suppress that header when needed.
Default behavior is preserved: if no opt-out is provided, requests still include ML-Agent-ID: nodejs.

### What Changed

1. Added conditional telemetry header logic in requester.js: 
    - ML-Agent-ID is now added only when disableTelemetryHeader is not true.
2. Added disableTelemetryHeader to allowed connection parameters in marklogic.js.
3. Documented the new connection option in createDatabaseClient JSDoc in marklogic.js.
4. Added TypeScript typing for disableTelemetryHeader in marklogic.d.ts.
5. Added README documentation for:
   - default telemetry header behavior,
   - what it discloses,
   - and how to opt out with disableTelemetryHeader in README.md.
6. Added tests for both behaviors:
   - JS unit test: header present by default and omitted when disabled in telemetry-header-test.js
   - TS runtime smoke test with the same assertions in telemetry-header-runtime.test.ts
   - TS compile-time coverage for new config option in basic-types.test.ts:58

### Why This Approach

1. Minimal surface-area change:
   - only gates a single header write,
   - no auth flow or request path behavior changes beyond the telemetry header.
2. Backward compatibility:
   - existing clients remain unchanged unless disableTelemetryHeader is explicitly set.
3. Clear security posture:
   - keeps telemetry by default for ecosystem consistency,
   - allows explicit opt-out for hardened environments.
4. Aligns with acceptance criteria:
   - implementation, docs, typings, and tests are all covered.

### Validation

1. Use Node 22 or higher.
2. Install dependencies: `npm install`
3. Run focused JavaScript test for header behavior: `npx mocha telemetry-header-test.js`
4. Compile TypeScript runtime tests: `npm run test:compile`
5. Run TypeScript runtime tests: `npm run test:typescript`

### Expected Results

1. test-basic telemetry test passes both cases:
   - Includes ML-Agent-ID by default
   - Omits ML-Agent-ID when disableTelemetryHeader is true
2. TypeScript compile completes with no errors.
3. TypeScript runtime suite executes successfully in Node 22+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,10 @@ const db = marklogic.createDatabaseClient({
   user:     'admin',
   password: 'admin',
   authType: 'DIGEST',
-  // enableGzippedResponses is optional and can be set to true in order to request MarkLogic to compress the response for better performance,
-    // the client will automatically decompress the response before it returns a value.
+  // Set to true to suppress the ML-Agent-ID header in each request to MarkLogic.
+  // See https://docs.progress.com/bundle/marklogic-server-monitor-12/page/topics/telemetry.html
+  disableTelemetryHeader: false,
+  // Optional; set to true to request compressed responses for improved performance. The client automatically decompresses returned values.
   enableGzippedResponses: true
 });
 
@@ -70,10 +72,9 @@ const db = marklogic.createDatabaseClient({
     authType: 'cloud',
     // basePath is optional.
     basePath: '/marklogic/test',
-    // accessTokenDuration (in seconds) is optional and can be used to customize the expiration of the access token.
+    // Optional (in seconds); customizes the expiration of the access token.
     accessTokenDuration: 10,
-    // enableGzippedResponses is optional and can be set to true in order to request MarkLogic to compress the response for better performance,
-    // the client will automatically decompress the response before it returns a value.
+    // Set to true to request compressed responses for improved performance. The client automatically decompresses returned values.
     enableGzippedResponses: true
 });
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -97,6 +99,16 @@ db.createCollection(
   });
 ```
 
+### Telemetry Header
```
rjrudin, line 102: Couple thoughts here. First, I don't think we want to establish a pattern of every new feature being documented in the `README` file. That file should typically be brief, providing an overview of the client and links to the documentation and other resources. This establishes a pattern that will produce an enormous `README` file over time. The right place for this information is in the existing Node Client docs. 

Second - I would provide a link to the MarkLogic telemetry docs in the type definition above - https://docs.progress.com/bundle/marklogic-server-monitor-12/page/topics/telemetry.html . And possibly in the config example above as well. I think saying that it enables or disables telemetry usage with a link to the telemetry docs is all we want.

README.md
```diff
@@ -58,6 +58,8 @@ const db = marklogic.createDatabaseClient({
   user:     'admin',
   password: 'admin',
   authType: 'DIGEST',
+  // disableTelemetryHeader is optional and can be set to true to suppress the ML-Agent-ID header.
```
rjrudin, line 61: I would make this and the next comment briefer by removing the name of the field, since it's obvious what it's referring too - e.g. `Optional; set to true to suppress inclusion of the ML-Agent-ID header in each request to MarkLogic.`.

rjrudin (approved):

rjdew-progress (approved):

Issue #1104, opened 2026-08-06
client.documents.write() fails on an array of documents


fgeorges:
I use the following module in `debug-write.mjs`, to write 2 documents in `Documents`:

```js
"use strict";

import ml from 'marklogic';

const client = ml.createDatabaseClient({
    host:     'localhost',
    user:     'admin',
    password: 'admin',
    port:     8000,
    authType: 'DIGEST',
});
client.setLogger('debug');

const docs = [
    {content: {hello: 'world'},       uri: '/test/hello.json', contentType: 'application/json'},
    {content: '<hello>world</hello>', uri: '/test/hello.xml',  contentType: 'application/xml'},
];

const res = await client.documents.write(docs).result();
console.dir(res);
```

I execute it like the following:

```sh
➜  time node debug-write.mjs
Debug: challenge request for /v1/ping
Debug: response with status 401 and true challenge for /v1/ping
Debug: authenticated request for /v1/documents
Debug: digest authentication
Debug: writing 2 parts
Debug: starting part 0
Debug: finished part 0
Debug: starting part 1
Debug: finished part 1
Error: socket hang up
Debug: deferred promise rejecting with 1 error messages
node:_http_client:782
    emitErrorEvent(req, new ConnResetException('socket hang up'));
                        ^

Error: socket hang up
    at Socket.socketOnEnd (node:_http_client:782:25)
    at Socket.emit (node:events:526:24)
    at endReadableNT (node:internal/streams/readable:1772:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  code: 'ECONNRESET'
}

Node.js v26.6.0
node debug-write.mjs  0.18s user 0.10s system 0% cpu 32.055 total
➜ 
```

There is more than 15s between these two lines:

```
Debug: finished part 1
Error: socket hang up
```

In `8000_AccessLog.txt`:

```
172.18.0.1 - - [06/Aug/2026:12:10:12 +0200] "HEAD /v1/ping HTTP/1.1" 401 0 - -
172.18.0.1 - admin [06/Aug/2026:12:10:44 +0200] "POST /v1/documents HTTP/1.1" 500 273 - -
```

In `8000_ErrorLog.txt`:

```
2026-08-06 12:10:44.590 Info: Status 500: XDMP-MULTIPART-BOUNDARY: xdmp:get-request-part-body("json") -- Ending of the boundary is incorrect
```

[...]

PR #1116, opened 2026-08-24
MLE-32051 Fix multipart body termination for Node.js v26 compatibility

jonmille:
## Summary

Fixes `client.documents.write()` failing with `ECONNRESET` / socket hang-up when writing an array of documents on Node.js v26.x.

Closes [GH #1104](https://github.com/marklogic/node-client-api/issues/1104) | Jira: [MLE-32051](https://progresssoftware.atlassian.net/browse/MLE-32051)

---

## Problem

When calling `client.documents.write([doc1, doc2, ...])` on Node.js v26, the HTTP request body is never properly terminated. MarkLogic receives the part content but not the closing multipart boundary (`--boundary--`), and responds with:

```
HTTP 500: XDMP-MULTIPART-BOUNDARY: xdmp:get-request-part-body("json")
         -- Ending of the boundary is incorrect
```

The client then receives an `ECONNRESET` after ~30 seconds.

**Root cause:** `multipart-stream` 2.0.1 uses `sandwich-stream` (a `Readable`) to assemble the multipart body. For each part it creates a `PassThrough` stream, writes headers + CRNL + body content as separate `.write()` calls, and adds it to `sandwich-stream`. `sandwich-stream`'s `_currentStreamOnReadable` calls `read()` to pull data from each sub-stream. On Node.js v26, `read()` returns only the **first write-chunk** from a PassThrough; the `'readable'` event does not re-fire for subsequent chunks, leaving them stranded. Because `_currentStreamOnEnd` never triggers, `sandwich-stream`'s `_pushTail()` / `push(null)` is never reached and the closing boundary is never emitted.

The bug affects **Node.js v26.6.0** (the reporter's version) and **v26.7.0**. Node.js v24.x is not affected.

---

## Fix

requester.js — bypasses `sandwich-stream` in both affected code paths:

- **`multipartRequester()` — `requestPartList` path** (`client.documents.write(array)`): detects whether any part carries stream content (`typeof content.pipe === 'function'`).
  - **All-synchronous content** (the v26 bug scenario — array of document objects): the multipart body is assembled directly as incremental `Buffer` writes to the request, avoiding `sandwich-stream` entirely.
  - **Stream content present** (binary writes, large documents, Data Services `ValueStream`): falls back to `multipart-stream` with `pipe()`. Stream-driven PassThroughs receive writes asynchronously after listeners are attached, so the Node.js v26 `'readable'` regression does not affect them.
- **`chunkedMultipartRequester()`**: non-stream (metadata) parts are written to the request as `Buffer` chunks; the final streaming content part is forwarded via `requestWriter.pipe(request, { end: false })` followed by the closing boundary and `request.end()`.

The `bindingParam` and `requestPartsProvider` paths in `multipartRequester()` retain `multipart-stream` with `pipe()` and an explicit `err => request.destroy(err)` error handler to properly tear down the socket on stream errors. 

---

## Testing

Tested manually with `repro.mjs` (see MLE-32051):

| Node.js | Before fix | After fix |
|---------|-----------|-----------|
| v24.19.0 | PASS | PASS |
| v26.6.0 | **FAIL** (ECONNRESET after 30s) | PASS (0.39s) |
| v26.7.0 | **FAIL** (ECONNRESET after 30s) | PASS (0.70s) |

MarkLogic server: `marklogic-server-ubi:11.3.6-ubi-2.2.6` 

Existing tests continue to pass with: `npx mocha test-basic/documents-core.js --timeout 0

[MLE-32051]: https://progresssoftware.atlassian.net/browse/MLE-32051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/requester.js
+++ b/lib/requester.js
@@ -477,14 +477,14 @@ function multipartRequester(request) {
       },
       body: form,
     });
-    multipartStream.on('data', chunk => request.write(chunk));
-    multipartStream.on('end', () => request.end());
+    multipartStream.on('error', err => request.destroy(err));
+    multipartStream.pipe(request);
   } else if (typeof requestPartsProvider === 'function') {
     // requestPartsProvider path: use multipart-stream (provider controls parts)
     const multipartStream = new Multipart(boundary);
     requestPartsProvider.call(operation, multipartStream);
-    multipartStream.on('data', chunk => request.write(chunk));
-    multipartStream.on('end', () => request.end());
+    multipartStream.on('error', err => request.destroy(err));
+    multipartStream.pipe(request);
   } else {
     // requestPartList path: all parts have synchronous content — build body directly
     // as a Buffer to avoid sandwich-stream's 'readable' event regression on Node.js v26+.
@@ -513,7 +513,8 @@ function multipartRequester(request) {
       operation.logger.debug('no part list to write');
     }
     bufs.push(Buffer.from(CRNL + '--' + boundary + '--'));
-    request.end(Buffer.concat(bufs));
+    for (let i = 0; i < bufs.length - 1; i++) { request.write(bufs[i]); }
+    request.end(bufs[bufs.length - 1]);
   }
 }
 function chunkedRequester(request) {
@@ -572,7 +573,8 @@ function chunkedMultipartRequester(request) {
           if (written > 0) { request.write(Buffer.from(CRNL + '--' + boundary + CRNL)); }
           Object.entries(headers).forEach(([k, v]) => request.write(Buffer.from(k + ': ' + v + CRNL)));
           request.write(Buffer.from(CRNL));
-          requestWriter.on('data', chunk => request.write(chunk));
+          requestWriter.on('error', err => request.destroy(err));
+          requestWriter.pipe(request, { end: false });
           requestWriter.on('end', () => {
             request.write(Buffer.from(CRNL + '--' + boundary + '--'));
             request.end();
--- a/lib/requester.js
+++ b/lib/requester.js
@@ -420,11 +420,13 @@ function multipartRequester(request) {
   const operation = this;
 
   const operationBoundary = operation.multipartBoundary;
-  const multipartStream = new Multipart((operationBoundary == null) ?
-        mlutil.multipartBoundary : operationBoundary);
+  const boundary = (operationBoundary == null) ? mlutil.multipartBoundary : operationBoundary;
+  const CRNL = '\r\n';
 
   const requestPartsProvider = operation.requestPartsProvider;
-  if(operation.bindingParam) {
+  if (operation.bindingParam) {
+    // bindingParam path: use multipart-stream (form-data body is a stream)
+    const multipartStream = new Multipart(boundary);
     const form = new formData();
     const bindingParam = operation.bindingParam;
     const query = bindingParam.query;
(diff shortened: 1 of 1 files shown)


lib/requester.js
RitaChen609, line 1: Can you add repro tests to the test suite in the repo?

RitaChen609:
@jonmille When I ran "npx mocha test-basic/bindingFromParam.js" on nodejs 26, 22 tests were failing.  Apparently, the PR's fix for nodejs 26 is incomplete. 

**FYI Copilot analysis:**

test-basic/bindingFromParam.js reproduces the same underlying multipart-boundary-termination bug from MLE-32051:
	  
`3 passing / 22 failing — all failures are 20000ms timeouts (hung write, never completed)`

Root cause match: db.rows.query(plan, options, bindingParam) sets operation.bindingParam in rows.js:256, which routes through the operation.bindingParam branch of multipartRequester() in requester.js:427. Looking at the current PR code, this branch was left unchanged — it still relies entirely on multipart-stream's addPart() (wrapping the form-data object as a single part), which the PR's own root-cause analysis identifies as the buggy code path on Node v26.

Why even the "no attachment" tests hang: addPart() always writes the part's headers (Content-Type, Accept) as separate synchronous .write() calls to a fresh PassThrough, then pipes the form stream into it — the header writes alone are enough to trigger the same "only first buffered chunk survives, no further readable" regression, so every bindingParam-based query hangs, not just ones with binary/stream attachments.

Why exactly 2 tests pass: 'test without column' and 'test with wrong column type;' never reach the network layer at all — they throw synchronously during op.fromParam(...) argument validation (caught in a try/catch), so no multipart request is ever sent.

This is a more severe gap than the "mixed array" issue — it means the PR's fix is incomplete not just for an edge case, but for an entire, commonly-used feature (Optic fromParam bindings) on Node v26. Every real bindingParam write is broken.

jonmille:
> @jonmille When I ran "npx mocha test-basic/bindingFromParam.js" on nodejs 26, 22 tests were failing. Apparently, the PR's fix for nodejs 26 is incomplete.
> 
> **FYI Copilot analysis:**
> 
> test-basic/bindingFromParam.js reproduces the same underlying multipart-boundary-termination bug from MLE-32051:
> 
> `3 passing / 22 failing — all failures are 20000ms timeouts (hung write, never completed)`
> 
> Root cause match: db.rows.query(plan, options, bindingParam) sets operation.bindingParam in rows.js:256, which routes through the operation.bindingParam branch of multipartRequester() in requester.js:427. Looking at the current PR code, this branch was left unchanged — it still relies entirely on multipart-stream's addPart() (wrapping the form-data object as a single part), which the PR's own root-cause analysis identifies as the buggy code path on Node v26.
> 
> Why even the "no attachment" tests hang: addPart() always writes the part's headers (Content-Type, Accept) as separate synchronous .write() calls to a fresh PassThrough, then pipes the form stream into it — the header writes alone are enough to trigger the same "only first buffered chunk survives, no further readable" regression, so every bindingParam-based query hangs, not just ones with binary/stream attachments.
> 
> Why exactly 2 tests pass: 'test without column' and 'test with wrong column type;' never reach the network layer at all — they throw synchronously during op.fromParam(...) argument validation (caught in a try/catch), so no multipart request is ever sent.
> 
> This is a more severe gap than the "mixed array" issue — it means the PR's fix is incomplete not just for an edge case, but for an entire, commonly-used feature (Optic fromParam bindings) on Node v26. Every real bindingParam write is broken.

I also identified that these tests were also related to the changes, and I confirmed they are passing now:
[documents-core.js] (array + stream writes)
[bindingFromParam.js]
[docColTypes-test.js]
[lockForUpdate-test.js]
[optic-remove.js]
[transformDoc-test.js]
[write-test.js]

lib/requester.js
RitaChen609: We don't need specific repro tests for this because the existing tests (test-basic/bindingFromParam.js, test-basic/optic-cts-param-test.js, etc) reproduce the same bug.

RitaChen609:
@jonmille Thanks for the bindingParam fix — confirmed bindingFromParam.js now passes on Node v26.7.0. While re-testing, found two more issues worth addressing, but I think these two edge cases (mixed-content arrays, stream-typed binding attachments) should not block the PR since the existing test cases all passed rather than regressions of the reported bug. We can create a new JIRA ticket to address them later.

Copilot drafted the following summary.

**1. requestPartList mixed-content arrays still hang on Node v26 (confirmed via test)**
client.documents.write([...]) with an array that mixes plain content (e.g. a JSON doc) and streamed content (e.g. fs.createReadStream() for a binary doc) still hangs/times out on Node v26.7.0.

Root cause: in multipartRequester(), hasStreamContent is true whenever any part in the array streams, which routes the entire array — including the plain parts — back through multipart-stream's addPart():

```
if (hasStreamContent) {
  const multipartStream = new Multipart(boundary);
  ...
  multipartStream.addPart({ headers: headers, body: content });
  ...
  multipartStream.pipe(request);
}
```

addPart() always writes each part's headers + CRNL synchronously to a fresh PassThrough before piping/ending the body, regardless of whether that specific part is a stream — this is the same write pattern that triggers the Node v26 'readable' regression described in the PR. So a batch containing even one streamed part drags every other (plain) part back into the broken path.

Suggested fix: apply the same technique already used for bindingParam and chunkedMultipartRequester() here — write non-stream parts as Buffers directly to request, and only pipe the part(s) whose content is an actual stream, instead of falling back to multipart-stream for the whole array whenever any single part streams.

Recommended test case: an array write with one plain-content document and one stream-content document in the same call, e.g.: run on Node v26.x. and add in test-basic/documents-core.js under an array with mixed plain and stream content 

```
db.documents.write([
  { uri: '...json', contentType: 'application/json', content: '{"key1":"value 1"}' },
  { uri: '...png',  contentType: 'image/png',        content: fs.createReadStream(binaryPath) }
])
```


**2. bindingParam.getBuffer() fix may throw for stream-typed attachments**
form.getBuffer() (in form-data) only handles Buffer/string entries:

```
if (Buffer.isBuffer(this._streams[i])) {
  dataBuffer = Buffer.concat([dataBuffer, this._streams[i]]);
} else {
  dataBuffer = Buffer.concat([dataBuffer, Buffer.from(this._streams[i])]); // throws for a stream
}
```

Nothing in rows.js/requester.js currently restricts bindingParam.attachments values to strings — if a caller passes an actual fs.createReadStream() (as opposed to a string) as an attachment value, form.append() wraps it as a DelayedStream, and getBuffer() will throw TypeError: The first argument must be of type string or an instance of Buffer... instead of streaming it, which previously worked via multipartStream.add({body: form}).pipe(request).

Suggested fix: either validate/reject non-string attachment values early with a clear error, or detect stream-typed attachments and fall back to the streaming path for that case.

Recommended test case: db.rows.query(plan, options, bindingParam) where an attachment value is fs.createReadStream(...) rather than a string, on Node v26.x, to confirm whether this is an actual supported/broken scenario.

RitaChen609 (approved):

rjdew-progress (approved):

